PR #25, opened 2015-08-11
Resolve only IPv4 addresses

dgadelha:
EOSERV does not totally support IPv6 (FIX for game.eoserv.net with IPv6-capable network)

### DIFF
--- a/EOLib/Net/AsyncClient.cs
+++ b/EOLib/Net/AsyncClient.cs
@@ -83,16 +83,20 @@ namespace EOLib.Net
 				IPAddress ip;
 				if (!IPAddress.TryParse(ipOrHostname, out ip))
 				{
-					IPHostEntry entry = Dns.GetHostEntry(ipOrHostname);
-					if (entry.AddressList.Length == 0)
-						return false;
+                    IPAddress[] ipv4Addresses = Array.FindAll(Dns.GetHostEntry(ipOrHostname).AddressList, a => a.AddressFamily == AddressFamily.InterNetwork);
 
-					ipOrHostname = entry.AddressList[0].ToString();
-				}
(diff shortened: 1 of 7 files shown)


ethanmoffat:
This change breaks the alpha blending feature for map objects and actors by setting the Reach profile and removing use of the _playerBlend BlendState object in EOMapRenderer. See below screenshot:
image

The view should be like so, note the player glows a little (inventory inconsistency is a separate issue):
image

"HiDef" is required for the blend state that is in place. As far as I know there is no reason any graphics card from the last 10 years should be unable to handle the limited set of requirements to render this game.

The following changes are fine: 
- Proper release attributes (4eef7c5)
- Desc some exceptions, change client version font and position (2c98b34) - Segoe UI does not display correctly on my system (changing to "Segoe UI Regular" works fine). I have no alternate reports of Microsoft Sans Serif not working properly on other systems, there may be a graphics driver issue or OS incompatibility if you're on Vista or earlier. I develop on Win8 but have confirmed it works on Win7. I will add this note to the issue.
- Resolve IPv4 addresses (adfa886) (this one I might update to require a configuration value, if I add dual-mode sockets)

Please take some time to format the source code for consistency: set the editor to use tabs, 4 spaces per tab, and align them with the current code. Also please group any fields/properties with other relevant existing fields/properties for future changes.

dgadelha:
Visual Studio 2015 :-1:

dgadelha:
Also, I didn't knew GitHub put all the commits I done into Pull Request, I just did it for 1 commit, the IPv6 one.

ethanmoffat:
Ahh ok. I'll look into merging just the one commit.

dgadelha:
nah, i'm doing it.

PR #28, opened 2015-08-12
My Changes

dgadelha:
- IPv4-only address
- Remove other debugging for Release Build


ethanmoffat:
Please amend the pull request - the first 5 commits in the pull request I would be fine with.

I don't want to have exceptions in main - that is only there for release and is commented out for a reason (it hides where the error actually is). I also don't want XNAControls.csproj as part of EndlessClient.sln - XNAControls is completely independent of EndlessClient.

I thought when I checked earlier today that the pull request originally only had the first 5 commits in it - did it get updated? I'm starting to get annoyed with git... :-/

dgadelha:
Well, I remade it so it had only 2, as you can see my description.
I put XNAControls to test, wasn't supposed to appear here, omg.

Issue #149, opened 2022-03-25
Shorts, Threes, and Ints aren’t unsigned


sorokya:
Noticed while testing this with reoserv that the client reads shorts, threes, and ints as signed values.

This became an issue when reoserv tried to send a a session id that was bigger than a signed short can hold.

I tried fixing this myself but it quickly became a huge change to over 100 files and broke a lot of stuff.

ethanmoffat:
Yeah this is a big problem. I also went through and tried to update it to use signed values and had the same issue, 100+ files changed quickly became a headache.

The actual problem is that because `NumberEncoderService` takes an `int`, it will sign extend `short`s that are greater than `short.MaxValue`. See: b80ff5ceea98ab0f1517a18f8c7a38e8dab4a280

What I ended up doing to resolve it is just casting to `ushort` client-side. It's an ugly hack but it works. 

Which session id specifically was causing the issue?

sorokya:
In reoserv I generate a random short value between 0 and the max2 value.

https://github.com/sorokya/reoserv/blob/async-rewrite/src/player/player.rs#L154

the error was coming up during the welcome message packet. The session id reoserv sent was too big.

ethanmoffat:
I'll see if I can reproduce it against reoserv. That commit I linked (b80ff5ceea98ab0f1517a18f8c7a38e8dab4a280) was working for me with larger player IDs on GameServer and etheos

sorokya:
Ah you’re right casting it back to unsigned does fix it. I’ll just do that for now in my branch. :P

ethanmoffat:
I didn't have that issue for shorts in my testing. I also added a line to `PacketBuilder` that casts the `short` to `ushort` prior to calling `EncodeNumber` with it so it doesn't get sign-extended during encoding: https://github.com/ethanmoffat/EndlessClient/blob/5fa705fa1c38f5961ad389a69399f2b84d27d624/EOLib/Net/PacketBuilder.cs#L70

I forgot I updated it in both these places, but it was working consistently for me on both etheos and gameserver

sorokya:
You're right that totally fixes it. No need for the overload :)

ethanmoffat:
When playing on BU, this problem is pretty apparent once you get over `TWO_BYTE_MAX` HP:
<img width="122" alt="image" src="https://user-images.githubusercontent.com/8301078/196003296-58d40dd7-641e-48c2-86ef-6beb7f0a7a18.png">

[...]

PR #150, opened 2022-03-26
Update session id logic

sorokya:
This branch updates the Session ID logic around Account, Character, Welcome, and Warp packets to closer match the original GameServer and reoserv.

There's a big issue around PacketBuilder.AddShort not letting you add a ushort though!

### DIFF
--- a/EOLib/Domain/Account/AccountActions.cs
+++ b/EOLib/Domain/Account/AccountActions.cs
@@ -93,7 +93,7 @@ namespace EOLib.Domain.Account
         public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters)
         {
             var createAccountPacket = new PacketBuilder(PacketFamily.Account, PacketAction.Create)
-                .AddShort(_playerInfoRepository.SessionID)
+                .AddShort((ushort)_playerInfoRepository.SessionID)
                 .AddByte(255)
                 .AddBreakString(parameters.AccountName)
                 .AddBreakString(parameters.Password)
--- a/EOLib/Domain/Account/AccountActions.cs
+++ b/EOLib/Domain/Account/AccountActions.cs
@@ -93,7 +93,7 @@ namespace EOLib.Domain.Account
         public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters)
         {
             var createAccountPacket = new PacketBuilder(PacketFamily.Account, PacketAction.Create)
-                .AddShort((short)_playerInfoRepository.SessionID)
+                .AddShort(_playerInfoRepository.SessionID)
                 .AddByte(255)
                 .AddBreakString(parameters.AccountName)
                 .AddBreakString(parameters.Password)
--- a/EOLib/Domain/Account/AccountActions.cs
+++ b/EOLib/Domain/Account/AccountActions.cs
@@ -71,8 +71,6 @@ namespace EOLib.Domain.Account
             var reply = (AccountReply)response.ReadShort();
             if (reply >= AccountReply.OK_CodeRange)
             {
-                _playerInfoRepository.SessionID = (short)reply;
-
                 // Based on patch: https://github.com/eoserv/eoserv/commit/80dde6d4e7f440a93503aeec79f4a2f5931dc13d
                 // Account may change sequence start depending on the eoserv build being used
                 // Official software always updates sequence number
@@ -90,10 +88,10 @@ namespace EOLib.Domain.Account
             return reply;
         }
 
-        public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters)
+        public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters, short sessionID)
         {
             var createAccountPacket = new PacketBuilder(PacketFamily.Account, PacketAction.Create)
-                .AddShort((ushort)_playerInfoRepository.SessionID)
+                .AddShort((ushort)sessionID)
                 .AddByte(255)
                 .AddBreakString(parameters.AccountName)
                 .AddBreakString(parameters.Password)
@@ -151,7 +149,7 @@ namespace EOLib.Domain.Account
 
         Task<AccountReply> CheckAccountNameWithServer(string accountName);
 
-        Task<AccountReply> CreateAccount(ICreateAccountParameters parameters);
+        Task<AccountReply> CreateAccount(ICreateAccountParameters parameters, short sessionID);
 
         Task<AccountReply> ChangePassword(IChangePasswordParameters parameters);
     }
--- a/EOLib/Domain/Account/AccountActions.cs
+++ b/EOLib/Domain/Account/AccountActions.cs
@@ -71,7 +71,7 @@ namespace EOLib.Domain.Account
             var reply = (AccountReply)response.ReadShort();
             if (reply >= AccountReply.OK_CodeRange)
             {
-                _playerInfoRepository.AccountCreateID = (ushort)reply;
+                _playerInfoRepository.SessionID = (short)reply;
 
                 // Based on patch: https://github.com/eoserv/eoserv/commit/80dde6d4e7f440a93503aeec79f4a2f5931dc13d
                 // Account may change sequence start depending on the eoserv build being used
@@ -93,7 +93,7 @@ namespace EOLib.Domain.Account
         public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters)
         {
             var createAccountPacket = new PacketBuilder(PacketFamily.Account, PacketAction.Create)
-                .AddShort((short)_playerInfoRepository.AccountCreateID)
+                .AddShort((short)_playerInfoRepository.SessionID)
                 .AddByte(255)
                 .AddBreakString(parameters.AccountName)
                 .AddBreakString(parameters.Password)
--- a/EOLib/Domain/Account/AccountActions.cs
+++ b/EOLib/Domain/Account/AccountActions.cs
@@ -91,7 +91,7 @@ namespace EOLib.Domain.Account
         public async Task<AccountReply> CreateAccount(ICreateAccountParameters parameters, short sessionID)
         {
             var createAccountPacket = new PacketBuilder(PacketFamily.Account, PacketAction.Create)
-                .AddShort((ushort)sessionID)
+                .AddShort(sessionID)
                 .AddByte(255)
                 .AddBreakString(parameters.AccountName)
                 .AddBreakString(parameters.Password)
--- a/EOLib/Domain/Login/ILoginRequestGrantedData.cs
+++ b/EOLib/Domain/Login/ILoginRequestGrantedData.cs
@@ -6,7 +6,7 @@ namespace EOLib.Domain.Login
 {
     public interface ILoginRequestGrantedData : ITranslatedData
     {
-        short PlayerID { get; }
+        short SessionID { get; }
         int CharacterID { get; }
 
         short MapID { get; }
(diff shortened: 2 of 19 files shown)


EOLib/Domain/Account/AccountActions.cs
```diff
@@ -71,7 +71,7 @@ public async Task<AccountReply> CheckAccountNameWithServer(string accountName)
             var reply = (AccountReply)response.ReadShort();
             if (reply >= AccountReply.OK_CodeRange)
             {
-                _playerInfoRepository.AccountCreateID = (ushort)reply;
```
ethanmoffat, line 74: I would rather have these properties (and associated parameters) named after which session ID they are rather than naming them all `SessionID`. I think it hurts readability, especially when there are multiple competing different types of session IDs, and multiple different properties across repositories for tracking said session IDs. I'd be ok with calling it `AccountCreateSessionID` if you really want to update the naming.

EOLib/Domain/Protocol/InitReply.cs
```diff
@@ -5,14 +5,15 @@ public enum InitReply
         ClientOutOfDate = 1,
         Success = 2,
         BannedFromServer = 3,
-        MapFile = 4,
```
ethanmoffat, line 8: Is this enum just wrong? I grabbed it from eoserv years ago, and have wondered why some of the packets used a `Char` for these while some used `Byte`, but that would explain it

EOLib/Net/PacketBuilder.cs
```diff
@@ -67,7 +67,12 @@ public IPacketBuilder AddChar(byte b)
 
         public IPacketBuilder AddShort(short s)
         {
-            return AddBytes(_encoderService.EncodeNumber((ushort)s, 2));
+            return AddBytes(_encoderService.EncodeNumber(s, 2));
+        }
+
+        public IPacketBuilder AddShort(ushort s)
```
ethanmoffat, line 73: See my comment on #149 - could you elaborate on the behavior this fixes? I wasn't seeing issues with shorts > 32767 as session IDs because the cast to `ushort` ensured the correct value was encoded.

I would prefer not to add this overload if it isn't needed

ethanmoffat (commented):
What are the results of testing these changes against etheos/eoserv/GameServer? Normally I'd assume it works but I'd like you to confirm because of the changes to the InitReply enum and reading data around it.

You can use `test.etheos.moffat.io:8078` for etheos, and `moffat.io:8079` for gameserver

EOLib/Domain/Account/AccountActions.cs
sorokya: I guess a name like ‘PregameSessionID’ might be better? It’s used for account create, character create, character delete, welcome agree, and welcome message packets.

I just called them all session_id in reoserv because only one could ever be in use at a time 🤷‍♂️

EOLib/Domain/Protocol/InitReply.cs
sorokya: That’s was my understanding from testing against GameServer. Eoserv actually adds the banned reply (as a chat) for warp map downloads haha

EOLib/Net/PacketBuilder.cs
sorokya: I’ll test it again. It seemed like it was overflowing or something. I kept getting a really huge number back on reoserv when the session id was bigger than a signed short can hold

sorokya:
> What are the results of testing these changes against etheos/eoserv/GameServer? Normally I'd assume it works but I'd like you to confirm because of the changes to the InitReply enum and reading data around it.
> 
> You can use `test.etheos.moffat.io:8078` for etheos, and `moffat.io:8079` for gameserver

I’ll try it out with both :)

EOLib/Domain/Protocol/InitReply.cs
ethanmoffat: Haha yeah I saw that when I checked it out...good find on this, it's a lot cleaner this way

EOLib/Domain/Account/AccountActions.cs
ethanmoffat: Alright, after checking things out a little more thoroughly here's where I've landed.

Character create/delete IDs aren't persisted anywhere; they're just used within the same controller method. Account create should be refactored to work the same way - there's really no need to have this property in a repository.

That means that `PlayerInfoRepository` should have a single ID - `PlayerID` - since that's what this actually is. I guess it could be called `SessionID` but I've been calling it `PlayerID` for 8+ years so I'd prefer to keep it with that name. This ID should be set during welcome.

Here's what I'd like for this PR:
1. Remove setting the session ID during network connection initialization, since it isn't needed anywhere else beyond completing the handshake
2. Keep the name of `PlayerID` instead of `SessionID`
3. Optional: refactor account create to return the account create ID from the call to `AccountActions.CheckAccountNameWithServer` in `AccountController`, following a similar pattern as character create/delete where the session ID is passed as a parameter to `AccountActions.CreateAccount`.
    - If you don't want to take this on, keep AccountCreateID as a separate property for now and I can refactor it later

sorokya: I've applied your suggestions :)

EOLib/Net/PacketBuilder.cs
sorokya: You're right casting to (ushort) before encoding totally fixes it. Not sure why the code didn't have that when I was testing.. Maybe I deleted it?? lmao

Works now without the overload :)

sorokya:
> > What are the results of testing these changes against etheos/eoserv/GameServer? Normally I'd assume it works but I'd like you to confirm because of the changes to the InitReply enum and reading data around it.
> > You can use `test.etheos.moffat.io:8078` for etheos, and `moffat.io:8079` for gameserver
> 
> I’ll try it out with both :)

Tested with eoserv and everything worked fine
I tried your test.etheos.moffat.io for etheos but the client couldn't connect.
GameServer is bombing on account create. I'll try to debug it further later

EOLib/Domain/Login/ILoginRequestGrantedData.cs
```diff
@@ -38,7 +38,7 @@ public interface ILoginRequestGrantedData : ITranslatedData
         short JailMap { get; }
         bool FirstTimePlayer { get; }
 
-        ILoginRequestGrantedData WithPlayerID(short playerID);
+        ILoginRequestGrantedData WithSessionID(short playerID);
```
sorokya, line 41: Need to rename argument 😵

ethanmoffat:
> > > What are the results of testing these changes against etheos/eoserv/GameServer? Normally I'd assume it works but I'd like you to confirm because of the changes to the InitReply enum and reading data around it.
> > > You can use `test.etheos.moffat.io:8078` for etheos, and `moffat.io:8079` for gameserver
> > 
> > 
> > I’ll try it out with both :)
> 
> Tested with eoserv and everything worked fine I tried your test.etheos.moffat.io for etheos but the client couldn't connect. GameServer is bombing on account create. I'll try to debug it further later

Try `etheos-test.westus3.azurecontainer.io:8078`...apparently my pipeline isn't deleting old dns entries so I've got a bunch of stale IP addresses on the `test.etheos.moffat.io` A record 😬

EndlessClient/Controllers/LoginController.cs
```diff
@@ -127,47 +129,47 @@ public async Task LoginToCharacter(ICharacter character)
                 if (unableToLoadMap || _fileRequestActions.NeedsFileForLogin(InitFileType.Map, _currentMapStateProvider.CurrentMapID))
                 {
                     gameLoadingDialog.SetState(GameLoadingDialogState.Map);
-                    if (!await SafeGetFile(async () => await _fileRequestActions.GetMapFromServer(_currentMapStateProvider.CurrentMapID)))
+                    if (!await SafeGetFile(async () => await _fileRequestActions.GetMapFromServer(_currentMapStateProvider.CurrentMapID, sessionID)))
```
ethanmoffat, line 132: For these lines, do this instead of creating an async lambda:
```csharp
if (!await SafeGetFile(() => _fileRequestActions.GetMapFromServer(_currentMapStateProvider.CurrentMapID, sessionID)))
    return;
```
It reduces the overhead in the async state machine when you return the task directly, since it is eventually await anyway

EOLib/Domain/Login/PlayerInfoRepository.cs
```diff
@@ -10,8 +10,6 @@ public interface IPlayerInfoRepository
 
         short PlayerID { get; set; }
```
ethanmoffat, line 11: Since you refactored the welcome packets to use the same pattern as everything else with session IDs, I think you can remove the `PlayerID` property here entirely. The only purpose of having it here was to pass it between the welcome request / welcome message handlers. I see you're using it now for the connection ID but I don't think that's used anywhere outside of the initial connection

sorokya: But this is where player ID is set. The ID sent from welcome reply is just another session ID so you kind of need to keep it here and set it from the connection action :)

sorokya: Is there somewhere else player ID is stored for other in game actions that require it?

ethanmoffat: I'm not seeing it used for any other in-game actions - everything else uses `Character.ID`.

I think it actually has been working by accident up until now. In welcome agree, I take the character matching the logged in character's name from the list of map characters and use that as the "main character". This has the side effect of overwriting the existing character ID (which is just character select ID) with the player ID, since this is what is sent from eoserv.

sorokya: It's kind of weird. The only place GameServer sends you your PlayerID is on the initial handshake.

But it does expect you to send back the CharacterID on Welcome_Msg
```
"Entering game"
client_packet(Welcome, Msg)
{
	three session_id
	int character_id
}
```

From what I can see in the protocol CharacterID is never used after entering the game. Looks like the server only expects it when
- Requesting login to a character
- deleting a character

So maybe there's no point of tracking CharacterID outside of the "LoggedIn" controller and we always just set the MainCharacter.ID to PlayerID from connecting? I guess we'd still have to track it in the PlayerInfoRepository right?

sorokya:
> > > > What are the results of testing these changes against etheos/eoserv/GameServer? Normally I'd assume it works but I'd like you to confirm because of the changes to the InitReply enum and reading data around it.
> > > > You can use `test.etheos.moffat.io:8078` for etheos, and `moffat.io:8079` for gameserver
> > > 
> > > 
> > > I’ll try it out with both :)
> > 
> > 
> > Tested with eoserv and everything worked fine I tried your test.etheos.moffat.io for etheos but the client couldn't connect. GameServer is bombing on account create. I'll try to debug it further later
> 
> Try `etheos-test.westus3.azurecontainer.io:8078`...apparently my pipeline isn't deleting old dns entries so I've got a bunch of stale IP addresses on the `test.etheos.moffat.io` A record 😬

Re-tested with GameServer and etheos. Both work as intended :)

EndlessClient/Controllers/LoginController.cs
sorokya: Done :)

EOLib/Domain/Login/PlayerInfoRepository.cs
sorokya: I went ahead and just left PlayerID in the repository and overwrite MainCharacter's ID with that when completing login

ethanmoffat: I think I'm going to clean this up further once it's merged. Right now `mainCharacter` is matched on the logged in character's name, it would be cleaner to match on the ID instead

ethanmoffat (approved):

PR #152, opened 2022-03-26
Select character with keyboard

sorokya:
Seemed like a fun feature to implement and get my hands dirty with the client.

No idea if this is the "correct" way to deal with adding keyboard handling to a component.

Let me know if there's a better way and I can fix it :)

### DIFF
--- a/EndlessClient/ControlSets/LoggedInControlSet.cs
+++ b/EndlessClient/ControlSets/LoggedInControlSet.cs
@@ -53,6 +53,7 @@ namespace EndlessClient.ControlSets
             _changePasswordButton = GetControl(currentControlSet, GameControlIdentifier.ChangePasswordButton, GetPasswordButton);
             _characterInfoPanels.AddRange(_characterInfoPanelFactory.CreatePanels(_characterSelectorProvider.Characters));
 
+            _allComponents.Add(new PreviousUserInputTracker(_endlessGameProvider, _userInputRepository));
             _allComponents.Add(new CurrentUserInputTracker(_endlessGameProvider, _userInputRepository));
             _allComponents.Add(_changePasswordButton);
             _allComponents.AddRange(_characterInfoPanels);
--- a/EndlessClient/ControlSets/LoggedInControlSet.cs
+++ b/EndlessClient/ControlSets/LoggedInControlSet.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Threading;
 using EndlessClient.Controllers;
 using EndlessClient.GameExecution;
+using EndlessClient.Input;
 using EndlessClient.UIControls;
 using EOLib.Domain.Login;
 using Microsoft.Xna.Framework;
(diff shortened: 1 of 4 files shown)


EndlessClient/ControlSets/LoggedInControlSet.cs
```diff
@@ -46,6 +53,7 @@ protected override void InitializeControlsHelper(IControlSet currentControlSet)
             _changePasswordButton = GetControl(currentControlSet, GameControlIdentifier.ChangePasswordButton, GetPasswordButton);
             _characterInfoPanels.AddRange(_characterInfoPanelFactory.CreatePanels(_characterSelectorProvider.Characters));
 
+            _allComponents.Add(new CurrentUserInputTracker(_endlessGameProvider, _userInputRepository));
```
ethanmoffat, line 56: One thing to test is that this doesn't cause a second `CurrentUserInputTracker` to be added to the game's components when transitioning to the `PlayingTheGame` game state. I think this is handled in `GameStateActions.RemoveOldComponents` but it would be good to double-check. If it works as expected then this should be fine.

EndlessClient/UIControls/CharacterInfoPanel.cs
```diff
@@ -168,6 +174,12 @@ protected virtual async Task DeleteButtonClick()
         protected virtual void DoUpdateLogic(GameTime gameTime)
         {
             _characterControl.Update(gameTime);
+
+            var keyboardState = _userInputProvider.CurrentKeyState;
+            if (keyboardState.IsKeyDown((Microsoft.Xna.Framework.Input.Keys)49+_characterIndex))
```
ethanmoffat, line 179: I think you want to check currentState.KeyUp and previousState.KeyDown for this, otherwise holding the key will spawn a bunch of tasks. The interlocked guard happens in `LoginButtonClick` so it won't stop the tasks from being started (and then immediately exiting).

ethanmoffat, line 179: Can you do `Keys.D1`+_characterIndex? I think C# is pretty flexible with adding values to enums in this way. 

Also, just add the using statement instead of fully qualifying the type here. FYI you can easily add the `using Microsoft.Xna.Framework.Input` via the keyboard shortcut CTRL+. in VS

ethanmoffat (changes requested):

EndlessClient/UIControls/CharacterInfoPanel.cs
sorokya: Will do :)

EndlessClient/ControlSets/LoggedInControlSet.cs
sorokya: I didn’t dig too deep into how the control set stuff worked. I assumed that the components were isolated to the login state and would clear when changing state. I’ll see if I can test that

EndlessClient/UIControls/CharacterInfoPanel.cs
sorokya: Done

sorokya: I switched to using IsKeyPressedOnce

EndlessClient/ControlSets/LoggedInControlSet.cs
sorokya: Confirmed that the components do get removed when changing control sets
devenv_PPXFlfnnMy
devenv_KJCq8JIpOC

ethanmoffat (approved):

PR #153, opened 2022-03-26
Reset chat on login

### LINKED ISSUE #151: Chat buffer isn't cleared between logins
When logging in multiple times without closing the client the server MOTD and global chat rules appear to duplicate for each login

EndlessClient_MeR7aQT7JU
EndlessClient_TMe6XWZsdS

Edit: Looks like chat messages persist as well

sorokya:
Simple change. Fixes #151 :)

### DIFF
--- a/EndlessClient/Controllers/LoginController.cs
+++ b/EndlessClient/Controllers/LoginController.cs
@@ -182,6 +182,8 @@ namespace EndlessClient.Controllers
                     return;
                 }
 
+                // TODO: This is a temporary workaround until the bug in AutomaticTypeMapper/Unity is resolved
+                // https://github.com/ethanmoffat/EndlessClient/issues/151#issuecomment-1079738889
                 ClearChat();
                 AddDefaultTextToChat();
 
--- a/EndlessClient/Controllers/LoginController.cs
+++ b/EndlessClient/Controllers/LoginController.cs
@@ -182,6 +182,7 @@ namespace EndlessClient.Controllers
                     return;
                 }
 
+                ClearChat();
                 AddDefaultTextToChat();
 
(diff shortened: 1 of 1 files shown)


EndlessClient/Controllers/LoginController.cs
```diff
@@ -229,6 +230,15 @@ private async Task<bool> SafeGetFile(Func<Task> operation)
             return await op.Invoke();
         }
 
+        private void ClearChat()
+        {
+            foreach(var chat in _chatRepository.AllChat.Values)
```
ethanmoffat, line 235: Do this via linq - `foreach (var chat in _chatRepository.AllChat.Values.Where(x => x.Count > 0)) { chat.Clear(); }`. You could also just do it unconditionally since there's only 6 or so chat collections anyway

EndlessClient/Controllers/LoginController.cs
```diff
@@ -229,6 +230,15 @@ private async Task<bool> SafeGetFile(Func<Task> operation)
             return await op.Invoke();
         }
 
+        private void ClearChat()
```
ethanmoffat, line 233: This workaround is fine, but add a todo note here to properly fix the `IResettable` issue with the repository

ethanmoffat (changes requested):

EndlessClient/Controllers/LoginController.cs
sorokya: Removed condition check :)

EndlessClient/Controllers/LoginController.cs
sorokya: Added TODO comment :)

ethanmoffat (approved):

PR #227, opened 2022-09-25
Minor UI positioning and text re-alignments

miou-gh:


EndlessClient/GameExecution/GameRunnerBase.cs
```diff
@@ -71,8 +72,14 @@ public virtual bool SetupDependencies()
                 if (string.Equals(arg, "--host") && i < _args.Length - 1)
                 {
                     var host = _args[i + 1];
-                    _registry.Resolve<IConfigurationRepository>()
-                        .Host = host;
+
+                    if (IPEndPoint.TryParse(host, out var endpoint)) {
```
ethanmoffat, line 76: Does IPEndPoint work with DNS names too or just IP addresses? `--host` should work with DNS names also

miou-gh: Yes, it works with hostnames as well.

ethanmoffat (approved):

PR #234, opened 2022-09-30
Update host arg to be separate host/port

ethanmoffat:
Remove IPEndpoint usage that didn't work with DNS hostnames.


ethanmoffat:
@miou-gh FYI I'm removing `IPEndPoint` usage in favor of a separate `--port` argument because `IPEndPoint` doesn't parse DNS hostnames and I rely on the existing behavior of `--host` for debugging purposes

PR #270, opened 2022-12-07
Update draw position for account creation menu labels

miou-gh:
Looks more accurate than before, that's about it.


ethanmoffat:
Closing this since this change is included in #271

PR #271, opened 2022-12-07
Fix scrollbar padding and dragging

miou-gh:


EndlessClient/UIControls/ScrollBar.cs
```diff
@@ -68,6 +68,7 @@ public class ScrollBar : XNAControl
             _scrollButton.SetParentControl(this);
 
             _totalHeight = DrawAreaWithParentOffset.Height;
+            this.ScrollToTop();
```
ethanmoffat, line 71: I think the `ScrollToTop` call here is a no-op since `ScrollOffset` is already set to `0`

If you want the `+1` offset then just add that to the `_scrollButton` position parameter in its constructor call.

EndlessClient/UIControls/ScrollBar.cs
```diff
@@ -88,7 +89,7 @@ public void ScrollToTop()
         {
             ScrollOffset = 0;
             var pixelsPerLine = (float)(scrollArea.Height - _scrollButton.DrawArea.Height * 2) / (_totalHeight - LinesToRender);
-            _scrollButton.DrawPosition = new Vector2(_scrollButton.DrawArea.X, _scrollButton.DrawArea.Height + pixelsPerLine * ScrollOffset);
+            _scrollButton.DrawPosition = new Vector2(_scrollButton.DrawArea.X, 1 + _scrollButton.DrawArea.Height + pixelsPerLine * ScrollOffset);
```
ethanmoffat, line 92: Since `ScrollOffset` is zero, the ` + pixelsPerLine * ScrollOffset` could be removed (it will always be zero). Oversight on my part when I first wrote it

EndlessClient/UIControls/ScrollBar.cs
```diff
@@ -141,10 +142,10 @@ private void arrowClicked(object btn, EventArgs e)
             }
 
             var pixelsPerLine = (float)(scrollArea.Height - _scrollButton.DrawArea.Height * 2) / (_totalHeight - LinesToRender);
-            _scrollButton.DrawPosition = new Vector2(_scrollButton.DrawPosition.X, _scrollButton.DrawArea.Height + pixelsPerLine * ScrollOffset);
-            if (_scrollButton.DrawPosition.Y > scrollArea.Height - _scrollButton.DrawArea.Height)
+            _scrollButton.DrawPosition = new Vector2(_scrollButton.DrawPosition.X, 1 + _scrollButton.DrawArea.Height + pixelsPerLine * ScrollOffset);
```
ethanmoffat, line 145: Can you explain why the `+1` and `-1` offsets were added to a bunch of the position calculations? Also curious if there's a cleaner way to do the offset without scattering these +1/-1 adjustments everywhere

EndlessClient/UIControls/ScrollBar.cs
```diff
@@ -160,12 +161,17 @@ private void OnScrollButtonDragged(object btn, EventArgs e)
             else if (y > scrollArea.Height - _scrollButton.DrawArea.Height)
                 y = scrollArea.Height - _scrollButton.DrawArea.Height;
 
-            _scrollButton.DrawPosition = new Vector2(0, y);
+            var dif = (CurrentMouseState.ScrollWheelValue - PreviousMouseState.ScrollWheelValue) / -160;
```
ethanmoffat, line 164: If 160 is the height of the scrollbar, different scrollbars have different heights (chat vs shop dialog vs account create dialog), so it can't be hard-coded

ethanmoffat: I also don't see where this variable is being used

ethanmoffat:
I'm not really sure what "smooth" dragging for scrollbar drag means, could you elaborate on the expected vs actual behavior? And the old vs new behavior?

PR #277, opened 2022-12-10
Label adjustments

CoderDanUK:
v29 official has no default text: 
image

old create account screen: 
image

new create account: 
image

v29 create account:
image


EndlessClient/ControlSets/LoginPromptControlSet.cs
```diff
@@ -110,7 +110,7 @@ private IXNATextBox GetLoginUserNameTextBox()
                 _textBoxCursor)
             {
                 MaxChars = 16,
-                DefaultText = "Username",
+                DefaultText = "",
```
ethanmoffat, line 113: You know, I _really_ like my default text, but the more I think about it it's kind of stupid lol

PR #282, opened 2022-12-11
Split inventory by server

CoderDanUK:
image

### DIFF
--- a/EndlessClient/HUD/Panels/ActiveSpellsPanel.cs
+++ b/EndlessClient/HUD/Panels/ActiveSpellsPanel.cs
@@ -43,7 +43,7 @@ namespace EndlessClient.HUD.Panels
         private readonly IESFFileProvider _esfFileProvider;
         private readonly ISpellSlotDataRepository _spellSlotDataRepository;
         private readonly ISfxPlayer _sfxPlayer;
-        private readonly IConfigurationRepository _configRepository;
+        private readonly IConfigurationProvider _configProvider;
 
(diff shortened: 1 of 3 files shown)


EndlessClient/HUD/Panels/ActiveSpellsPanel.cs
```diff
@@ -559,11 +563,11 @@ protected override void Dispose(bool disposing)
         }
 
         [SupportedOSPlatform("Windows")]
-        private static RegistryKey TryGetCharacterRegistryKey(string accountName, string characterName)
+        private static RegistryKey TryGetCharacterRegistryKey(string accountName, string characterName, string host)
         {
             using RegistryKey currentUser = Registry.CurrentUser;
 
-            var pathSegments = $"Software\\EndlessClient\\{accountName}\\{characterName}\\spells".Split('\\');
+            var pathSegments = $"Software\\EndlessClient\\{host}\\{accountName}\\{characterName}\\spells".Split('\\');
```
ethanmoffat, line 570: Registry is only used to load existing item layouts which are overwritten by whatever is in an ini file (if it exists). Registry key path should remain unchanged (for this and for items panel)

ethanmoffat (changes requested):

ethanmoffat (approved):
This looks good; for the future try to avoid force pushing/rewriting history on existing PRs, it screws up GitHub's PR tool for "changes since your last review"

PR #283, opened 2022-12-11
fixes issue #222 clear walk path after attack request

### LINKED ISSUE #222: Attacking after clicking to walk doesn't stop the walk
The walk path should be cleared if an attack is attempted after clicking to walk

CoderDanUK:
https://user-images.githubusercontent.com/464887/206895631-fbe531c1-f24b-4a7b-894c-e560067c7239.mp4

Fixes #222


EndlessClient/Rendering/Character/CharacterAnimator.cs
```diff
@@ -432,6 +431,11 @@ private CharacterRenderProperties FaceTarget(MapCoordinate characterCoord, MapCo
             return rp;
         }
 
+        public void ClearWalkPath()
+        {
+            _walkPath.Clear();
```
ethanmoffat, line 436: Instead of exposing details of `_walkPath` outside of `CharacterAnimator`, how about clearing the walk path in the `Attack` call of `CharacterAnimator`? Is there a reason that wouldn't work that I'm not thinking of?

CoderDanUK: I think you're right I'll have a look into this 😊

PR #284, opened 2022-12-11
Fixes issue #193 Jail Protection

### LINKED ISSUE #193: Jail protection
The following restrictions should be implemented for the jail map:
- You cannot drop items in jail (message box)
- You cannot trade in jail (message box)
- You cannot use global in jail (status message)

CoderDanUK:
+ Status message for using global chat
+ Status message for trading in jail

+ Protection against "Sending" a message from jail
+ Protection from using "~" in jail

+ Dialog box for trading in jail

image

image

Fixes #193


EOLib/Domain/Map/CurrentMapStateRepository.cs
```diff
@@ -46,6 +46,8 @@ public interface ICurrentMapStateProvider
 
         short JailMapID { get; }
 
+        bool IsJail { get; }
```
ethanmoffat, line 49: Even if it's a readonly propery, add this to `ICurrentMapStateRepository` too for consistency

EOLib.Graphics/NativeGraphicsLoader.cs
```diff
@@ -23,14 +23,14 @@ public IImage LoadGFX(GFXTypes file, int resourceValue)
             }
             catch (ArgumentException)
             {
-#if DEBUG
+#if DEBUG && THROW_ON_TEXTURE_FAILURE
```
ethanmoffat: I _want_ this to throw by default if a texture is missing.
a) Main clone is currently using fanfic textures. If one is missing, updating those is an exercise for the user. I'll support them eventually but not yet
b) Connecting to a server with custom assets (e.g. BU) is also an exercise for the user

I don't want potential gfx loading issues to be hidden. In release mode it's fine to not crash, in debug mode it should be obvious that something is wrong

EndlessClient/Controllers/ChatController.cs
```diff
@@ -85,6 +90,15 @@ public void SelectChatTextBox()
             _chatTextBoxActions.FocusChatTextBox();
         }
 
+        public void ClearAndWarnIfJailAndGlobal()
+        {
+            if (!_currentMapStateProvider.IsJail) return;
```
ethanmoffat, line 95: Combine these `if`s to a single `if` statement

EndlessClient/HUD/Controls/HudControlsFactory.cs
```diff
@@ -416,6 +416,7 @@ private ChatTextBox CreateChatTextBox()
             };
             chatTextBox.OnEnterPressed += (_, _) => _chatController.SendChatAndClearTextBox();
             chatTextBox.OnClicked += (_, _) => _chatController.SelectChatTextBox();
+            chatTextBox.OnTextChanged += (_, _) => _chatController.ClearAndWarnIfJailAndGlobal();
```
ethanmoffat, line 419: I'm not familiar with this behavior, is this what the vanilla client does?

EOLib.Graphics/NativeGraphicsLoader.cs
CoderDanUK: Ah apologies I thought I removed this from the pr, I checked it in by mistake. It was so that I could experience bald missing graphics on clone without crashing 💀

EndlessClient/HUD/Controls/HudControlsFactory.cs
CoderDanUK: It is yeah, I can get a video of it if you like. Although... Client does go one extra and appears to check the message box on a regular interval rather than on keypress...

For example, this implementation will allow endlessclient to type a global message in aeven, walk in to jail and the message will still be there. However, when they try to send, they will be prevented from being able to send the message anyway. 

In the vanilla client, if you type a global message then walk in to jail, the client has a small delay and removes the message within a couple hundred microseconds it seems. 

The other thing I could add to this pr is this same chatbox check and run that check on map warp too.

EOLib/Domain/Map/CurrentMapStateRepository.cs
CoderDanUK: done

EndlessClient/Controllers/ChatController.cs
CoderDanUK: done

EndlessClient/HUD/Controls/HudControlsFactory.cs
ethanmoffat: Keeping parity with the original client is important to me for this project, so as long as the original client does it, I'm fine with it. I just haven't played around with the behavior and wasn't able to test it when I was reviewing.

For the map change chat box clear, that would be nice to have. You can take a look at `MapChangedActions`, that's probably where you'd want to add it if the map is the jail map.

CoderDanUK: this is vanilla with some trash quality https://youtu.be/0qw7OS7Pi_c

PR #287, opened 2022-12-12
Implement Player Ghosting

DanDecrypted:


EOLib/Domain/Character/WalkValidationActions.cs
```diff
@@ -16,6 +24,10 @@ public class WalkValidationActions : IWalkValidationActions
         private readonly ICurrentMapStateProvider _currentMapStateProvider;
         private readonly IUnlockDoorValidator _unlockDoorValidator;
 
+        private DateTime? _startedGhosting = null;
+        private Character _currentlyGhosting = null;
```
ethanmoffat, line 28: One thing I'm not big on is storing state in this class directly. My ideal goal was to have state represented by repository/provider(s) throughout the code base, but there are a number of services that also have their own private state (see: `Characteranimator`). With the `Actions` layer though I would like to avoid it; I think it makes sense to introduce a repository for these objects.

EndlessClient/Input/UnwalkableTileActions.cs
```diff
@@ -74,26 +74,9 @@ public IReadOnlyList<UnwalkableTileAction> GetUnwalkableTileActions(IMapCellStat
...
-            //            EOResourceID.STATUS_LABEL_KEEP_MOVING_THROUGH_PLAYER);
-            //        if (_startWalkingThroughPlayerTime == null)
-            //            _startWalkingThroughPlayerTime = DateTime.Now;
-            //        else if ((DateTime.Now - _startWalkingThroughPlayerTime.Value).TotalSeconds > 5)
-            //        {
-            //            _startWalkingThroughPlayerTime = null;
-            //            goto case TileInfoReturnType.IsTileSpec;
-            //        }
-            return UnwalkableTileAction.Character;
+            return cellState.Warp.Match(
```
ethanmoffat, line 77: Is there a reason you did this in `WalkValidationActions` rather than here in `UnwalkableTileActions`?

I think my plan for this feature was to `return UnwalkableTileActions.Character` here and have the `ArrowKeyController` handle it accordingly (start ghosting, walk through after X seconds). IMO handling ghosting doesn't really belong in `WalkValidationActions` because it's a bit beyond the scope of validating if a walk can happen

EOLib/Domain/Character/WalkValidationActions.cs
```diff
@@ -27,10 +39,31 @@ public class WalkValidationActions : IWalkValidationActions
...
+            if (_currentlyGhosting == null)
+                return;
+
+            var mc = _characterProvider.MainCharacter;
+            var playersXDiff = Math.Abs(mc.RenderProperties.MapX - _currentlyGhosting.RenderProperties.MapX);
+            var playersYDiff = Math.Abs(mc.RenderProperties.MapY - _currentlyGhosting.RenderProperties.MapY);
+            var playersAreNextToEachother =  playersXDiff == 1 || playersYDiff == 1;
+            var playersAreOnTopOfEachother = playersXDiff == 0 && playersYDiff == 0;
+
+            var timerHasBeenTooLong = _finishedGhosting.HasValue ? (DateTime.Now - _finishedGhosting.Value).TotalSeconds > 2 : false;
```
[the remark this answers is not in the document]
ethanmoffat: This could also be reduced to `_finishedGhosting.HasValue && (DateTime.Now - _finishedGhosting.Value).TotalSeconds > 2`

EOLib/Domain/Character/WalkValidationActions.cs
CoderDanUK: This has now been moved into a `GhostingRepository`

EOLib/Domain/Character/WalkValidationActions.cs
```diff
@@ -16,6 +24,10 @@ public class WalkValidationActions : IWalkValidationActions
         private readonly ICurrentMapStateProvider _currentMapStateProvider;
         private readonly IUnlockDoorValidator _unlockDoorValidator;
 
+        private DateTime? _startedGhosting = null;
```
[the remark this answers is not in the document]
CoderDanUK: This greatly simplifies the logic too because you can do Timer.Start() and Timer.Stop() for finished trying to ghost

EndlessClient/Controllers/ArrowKeyController.cs
```diff
@@ -101,24 +110,35 @@ private void FaceOrAttemptWalk(EODirection direction)
...
-            else
-            {
-                _characterAnimationActions.StartWalking(Option.None<MapCoordinate>());
+            _walkValidationActions.ClearGhostingCacheIfNeeded();
 
-                var coordinate = new MapCoordinate(
-                    _characterProvider.MainCharacter.RenderProperties.MapX,
-                    _characterProvider.MainCharacter.RenderProperties.MapY);
-                _spikeTrapActions.HideSpikeTrap(coordinate);
+            var walkValidationResult = _walkValidationActions.CanMoveToDestinationCoordinates();
```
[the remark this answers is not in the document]
CoderDanUK: Well that was part of the reason for implementing the _finishedGhosting time being +1 second after the ghosting threshold, to give it that second window to actually go through with the ghost walk. You are right though that in this current implementation, if you spam an arrow key fast enough you will go through the player so some more thought needs to be had around this, but I like the progress of using a ghosting repository to group the logic together nicely - so good call on that!

EndlessClient/Rendering/ContextMenuRenderer.cs
```diff
@@ -122,6 +126,7 @@ private enum MenuAction
         public override void Initialize()
         {
             base.Initialize();
+            _sfxPlayer.PlaySfx(SoundEffectID.ButtonClick);
```
ethanmoffat, line 129: I'd avoid a `_sfxPlayer` call here, the paradigm for `Initialize()` is really more for setting up the control. Maybe this belongs in the `ContextMenuRendererFactory`?

ethanmoffat: Or even better, where the factory method gets called. Something about having an SFX plays with every call to `Initialize` on a class feels wrong to me, I'd rather have it tied to an action instead

EOLib/Domain/Character/GhostingRepository.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using AutomaticTypeMapper;
+
+namespace EOLib.Domain.Character
+{
+
+    public interface IGhostingRepository : IResettable
```
ethanmoffat, line 9: This is mixing the concepts of `Repository` and `Service`

1. Services are stateless* classes that do a thing
2. Repositories or Providers are classes that store a thing (mapped to the same implementation in IoC, provider is read-only, repository is read/write)
3. Actions coordinate between services and repositories/providers. Data from repositories/providers is passed to services, operated on, and then updated in said repositories/providers accordingly.

So for this class: if you want a repository, break out the state into something with get/setters. Inject a provider if the consumer only needs read access, otherwise inject a repository. I think the operations are probably suited to `WalkValidationActions`, but the state should be stored in a provider/repository.

*Sometimes they aren't stateless, but the goal is that they should be. `InventoryService` is a good example of a service that doesn't hold any state and just does a thing. Same with `NumberEncoderService`.

EndlessClient/Input/UnwalkableTileActions.cs
ethanmoffat: > this could perhaps be in a better location around the arrow key handler or something perhaps

That's basically what I'm suggesting. Do the handling code at the `ArrowKeyController` level. WalkValidationActions doesn't really change, just indicates that a character is there, and if you get `Character` for more than {x} seconds the controller lets you pass through.

I haven't played around with this yet though - so maybe there's something I'm not thinking of as to why it doesn't work.

EndlessClient/Rendering/ContextMenuRenderer.cs
ethanmoffat: Resolved in split-out PR for SFX stuff

EOLib/Domain/Character/GhostingRepository.cs
ethanmoffat: Resolved via local refactor (to be pushed)

EndlessClient/Input/UnwalkableTileActions.cs
ethanmoffat: This seems fine now that I've played around with it more. The way I've split the data vs methods out from the existing `GhostingRepository` it makes sense to keep this contained within `WalkValidationActions`.

EndlessClient/Controllers/ArrowKeyController.cs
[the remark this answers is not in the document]
ethanmoffat: Fixed in local commit

ethanmoffat (approved):

PR #291, opened 2022-12-19
Fix stat display on login

sorokya:
Apparently wis/int are flipped just for the welcome packet.. So dumb lol


ethanmoffat:
holy shit

PR #341, opened 2024-04-16
Out of drop range message

AstralSkies:
- Fix out-of-range item drop handling with message and icon.

image

### DIFF
--- a/EndlessClient/Controllers/InventoryController.cs
+++ b/EndlessClient/Controllers/InventoryController.cs
@@ -49,6 +49,7 @@ namespace EndlessClient.Controllers
         private readonly IItemTransferDialogFactory _itemTransferDialogFactory;
         private readonly IEOMessageBoxFactory _eoMessageBoxFactory;
         private readonly IChatRepository _chatRepository;
+        private readonly ILocalizedStringFinder _localizedStringFinder;
 
         public InventoryController(IItemActions itemActions,
                                    IInGameDialogActions inGameDialogActions,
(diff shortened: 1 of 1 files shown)


EndlessClient/Controllers/InventoryController.cs
```diff
@@ -230,7 +234,10 @@ public void DropItem(EIFRecord itemData, InventoryItem inventoryItem)
             }
             else if (validationResult == ItemDropResult.TooFar)
             {
+                var message = "Out of drop range.";
```
ethanmoffat, line 237: Inject `ILocalizedStringFinder` and use that to look up `EOResourceID.STATUS_LABEL_ITEM_DROP_OUT_OF_RANGE`.

Looks like `"System"` isn't in the EDFs so that's fine to leave hard-coded.

ethanmoffat: Usage example from recent `ADMININTERACT` change here: https://github.com/ethanmoffat/EndlessClient/commit/5d02996856f973075188f78676c584667786a217#diff-9a0776e511801824a93b00f7f2f9b499cea6fed37165d926f01058bed1213246R24

AstralSkies: Sure, I'll do that. My bad.

AstralSkies: If anyone can check this on a Linux setup, it would be greatly appreciated.  Unable to verify or debug the Linux build issues right now.

ethanmoffat: This might be a build agent issue, I'll look into it. Not sure why it's failing. Your changes shouldn't impact it.

Issue #319, opened 2023-05-25
Missing sound effect when giving "safety comment" for dropping a large amount of gold


ethanmoffat:
In the vanilla client, the sound effect `Login` (also `ServerMessage`) plays when the dialog is shown

AstralSkies:
If I remember correctly, the sound effect is applied once along with the safety message.

ethanmoffat:
You're correct, the dialog/sfx is only shown once

PR #343, opened 2024-04-20
Resolved: Missing Sound Effect on Gold Drop Warning #319

AstralSkies:
- Added "Login" sound effect to gold drop warning dialog.
- Ensured single occurrence of warning message and sound, aligning with Vanilla EO.

### DIFF
--- a/EndlessClient/Controllers/InventoryController.cs
+++ b/EndlessClient/Controllers/InventoryController.cs
@@ -23,6 +23,7 @@ using EOLib.Localization;
 using System;
 using System.Linq;
 using XNAControls;
+using System.Diagnostics;
 
 namespace EndlessClient.Controllers
 {
@@ -50,6 +51,7 @@ namespace EndlessClient.Controllers
         private readonly IEOMessageBoxFactory _eoMessageBoxFactory;
         private readonly IChatRepository _chatRepository;
         private readonly ILocalizedStringFinder _localizedStringFinder;
+        private bool _goldWarningShown = false;
 
         public InventoryController(IItemActions itemActions,
                                    IInGameDialogActions inGameDialogActions,
@@ -238,7 +240,7 @@ namespace EndlessClient.Controllers
             else if (validationResult == ItemDropResult.TooFar)
             {
                 var localizedMessage = _localizedStringFinder.GetString(EOResourceID.STATUS_LABEL_ITEM_DROP_OUT_OF_RANGE);
-                var chatData = new ChatData(ChatTab.System, "System",localizedMessage, ChatIcon.DotDotDotDot);
+                var chatData = new ChatData(ChatTab.System, "System", localizedMessage, ChatIcon.DotDotDotDot);
                 _statusLabelSetter.SetStatusLabel(EOResourceID.STATUS_LABEL_TYPE_WARNING, EOResourceID.STATUS_LABEL_ITEM_DROP_OUT_OF_RANGE);
                 _chatRepository.AllChat[ChatTab.System].Add(chatData);
             }
@@ -350,8 +352,8 @@ namespace EndlessClient.Controllers
         }
 
         private void DoItemDrop(EIFRecord itemData, InventoryItem inventoryItem, Action<int> dropAction,
-                                ItemTransferDialog.TransferType transferType = ItemTransferDialog.TransferType.DropItems,
-                                EOResourceID message = EOResourceID.DIALOG_TRANSFER_DROP)
+                         ItemTransferDialog.TransferType transferType = ItemTransferDialog.TransferType.DropItems,
+                         EOResourceID message = EOResourceID.DIALOG_TRANSFER_DROP)
         {
             if (inventoryItem.Amount > 1)
             {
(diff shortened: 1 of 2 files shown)


EndlessClient/Dialogs/Factories/ItemTransferDialogFactory.cs
```diff
@@ -44,7 +46,19 @@ public ItemTransferDialog CreateItemTransferDialog(string itemName, ItemTransfer
                 transferType,
                 totalAmount,
                 message);
-            dlg.DialogClosing += (_, _) => _sfxPlayer.PlaySfx(SoundEffectID.DialogButtonClick);
+
+            dlg.DialogClosing += (_, _) =>
```
ethanmoffat, line 50: `ItemTransferDialogFactory` shouldn't have/need any changes, the sound effect should be added to the controller where the `warningMsg` dialog is created and shown.

This class should stay generic to creating a dialog for transferring items somewhere, not have special logic for gold drop.

EndlessClient/Dialogs/Factories/ItemTransferDialogFactory.cs
```diff
@@ -44,7 +46,19 @@ public ItemTransferDialog CreateItemTransferDialog(string itemName, ItemTransfer
...
+            {
+                if (itemName == "Gold" && !_goldSoundPlayed)
```
ethanmoffat, line 52: When you move this logic to the controller, you can add the sound effect to when the "safety comment" message dialog is opened rather than when the item transfer is closed. I think that makes a bit more sense.

ethanmoffat (changes requested):

EndlessClient/Dialogs/Factories/ItemTransferDialogFactory.cs
AstralSkies: Sure, I'll do that 👍

AstralSkies:
Hi Ethan,

Quick update: the recent unusual commit history was due to my sister experimenting with Git on my project. I’ve cleaned up the history and addressed all requested changes in the latest commit (96cb547). The PR is now ready for your review again.

EndlessClient/Dialogs/Factories/ItemTransferDialogFactory.cs
```diff
@@ -44,13 +44,21 @@ public ItemTransferDialog CreateItemTransferDialog(string itemName, ItemTransfer
                 transferType,
                 totalAmount,
                 message);
-            dlg.DialogClosing += (_, _) => _sfxPlayer.PlaySfx(SoundEffectID.DialogButtonClick);
+
+            if (playDefaultSound)
```
ethanmoffat, line 48: It looks like the official client doesn't play any sound for the "ok" button. I checked "drop", "junk", "buy", and "sell". I'm good with either:

1) Only play the `DialogButtonClick` sound on "cancel" button press
2) Play `DialogButtonClick` on either button. Play `DialogButtonClick` and safety comment message together

I don't really care too much about the sounds overlapping in this case. I'd rather avoid having a branching arg in the method signature for which sound is played.

That means the options for this file are:
1) Undo all the changes, then check which button was pressed and only play the default sfx if the button is "cancel"
2) Undo all the changes and leave this file untouched

EndlessClient/Controllers/InventoryController.cs
```diff
@@ -50,6 +51,9 @@ public class InventoryController : IInventoryController
         private readonly IEOMessageBoxFactory _eoMessageBoxFactory;
         private readonly IChatRepository _chatRepository;
         private readonly ILocalizedStringFinder _localizedStringFinder;
+        private readonly ISfxPlayer _sfxPlayer;
+
+        private bool _goldWarningShown = false;
```
ethanmoffat, line 56: Now that this class contains state, you will need to change `[AutoMappedType]` to `[AutoMappedType(IsSingleton = true)]`

EndlessClient/Controllers/InventoryController.cs
```diff
@@ -348,29 +354,41 @@ public void TradeItem(EIFRecord itemData, InventoryItem inventoryItem)
...
-                                ItemTransferDialog.TransferType transferType = ItemTransferDialog.TransferType.DropItems,
-                                EOResourceID message = EOResourceID.DIALOG_TRANSFER_DROP)
+                        ItemTransferDialog.TransferType transferType = ItemTransferDialog.TransferType.DropItems,
+                        EOResourceID message = EOResourceID.DIALOG_TRANSFER_DROP)
         {
+            bool playDefaultSound = true;
+
             if (inventoryItem.Amount > 1)
             {
+                if (inventoryItem.ItemID == 1 && inventoryItem.Amount > 10000 && transferType == ItemTransferDialog.TransferType.DropItems)
```
ethanmoffat, line 365: If you want to break the condition out of the if statement, do something like

`var isLargeGoldItemDrop = inventoryItem.ItemID == 1 && inventoryItem.Amount > 10000 && transferType == ItemTransferDialog.TransferType.DropItems;`

I think this should also move closer to where it's used since I'm asking you to remove the `playDefaultSound` parameter from the item transfer dialog factory too.

ethanmoffat (changes requested):

EndlessClient/Dialogs/Factories/ItemTransferDialogFactory.cs
AstralSkies: I'd rather go for 1) as that's what I initially made me include playDefaultSound: 

It would mean the inclusion of a new using statement along with 
image

Would this be okay?

Best,
Stephen

EndlessClient/Controllers/InventoryController.cs
AstralSkies: image

EndlessClient/Controllers/InventoryController.cs
AstralSkies: image

ethanmoffat (approved):

ethanmoffat:
LGTM, thanks for sticking with it through the feedback cycle. Much appreciated!

PR #347, opened 2024-04-25
Reoserv Fixes

sorokya:
This branch contains all the fixes for the bugs listed here:
https://docs.google.com/spreadsheets/d/1knlYz-HFolGtnz_yLi7ZrTkYN6AZKkNV6jO9Lri7BBg/edit#gid=0

Let me know if you have any questions :)

### DIFF
--- a/EOLib/PacketHandlers/Players/PlayersAgreeHandler.cs
+++ b/EOLib/PacketHandlers/Players/PlayersAgreeHandler.cs
@@ -18,6 +18,7 @@ namespace EOLib.PacketHandlers.Players
     [AutoMappedType]
     public class PlayersAgreeHandler : InGameOnlyPacketHandler
     {
+        private readonly IPacketTranslator<PlayersAgreeData> _playersAgreeTranslator;
         private readonly ICharacterRepository _characterRepository;
         private readonly ICurrentMapStateRepository _mapStateRepository;
         private readonly ICharacterFromPacketFactory _characterFromPacketFactory;
@@ -33,7 +34,8 @@ namespace EOLib.PacketHandlers.Players
                                      ICurrentMapStateRepository mapStateRepository,
                                      ICharacterFromPacketFactory characterFromPacketFactory,
                                      IEIFFileProvider eifFileProvider,
-                                     IEnumerable<IEffectNotifier> effectNotifiers)
+                                     IEnumerable<IEffectNotifier> effectNotifiers,
+                                     IPacketTranslator<PlayersAgreeData> playersAgreeTranslator)
             : base(playerInfoProvider)
         {
             _characterRepository = characterRepository;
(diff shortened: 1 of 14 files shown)


EOLib/PacketHandlers/Players/PlayersAgreeHandler.cs
```diff
@@ -33,37 +34,22 @@ public class PlayersAgreeHandler : InGameOnlyPacketHandler
...
-                    notifier.NotifyWarpEnterEffect(character.ID, anim);
-            }
-
-            if (packet.ReadByte() != 255)
-                throw new MalformedPacketException("Missing 255 byte after the warp animation for player enter map handler", packet);
-
-            // 0 for NPC, 1 for player. In eoserv it is never 0.
-            if (packet.ReadChar() != 1)
-                throw new MalformedPacketException("Missing '1' char after warp animation for player enter map handler. Are you using a non-standard version of EOSERV?", packet);
+            var data = _playersAgreeTranslator.TranslatePacket(packet);
```
ethanmoffat, line 51: This breaks eoserv compatibility. The PLAYERS_AGREE packet that eoserv sends is just...wrong. But because of the way the chunked reader works, it handles the wrong data correctly. EndlessClient won't have this functionality until eolib-dotnet is done.

To repro, try warping with another character to the map the main character is on.

ethanmoffat (commented):
I'm hesitant to merge this because it breaks compatibility with eoserv on player enter map, and I don't see a good way to support eoserv's weird way of doing it while supporting the actual packet formats without a proper chunked reader.

Everything else looks fine, but when I finish eolib-dotnet and migrate that'll address all these issues anyway, while still supporting both emulators.

EOLib/PacketHandlers/Players/PlayersAgreeHandler.cs
Cirras: For more information on the gnarly `PLAYERS_AGREE` data sent by the official server (and blindly mimicked by EOSERV), see the eo-protocol docs.

https://github.com/Cirras/eo-protocol/blob/master/docs/chunks.md#a-practical-example

sorokya: @ethanmoffat also realized the solution still isn't correct because `_characterFromPacketFactory.CreateCharacter` doesn't check for the WarpAnimation byte.

Idk where the proper place to put this would be. On Character itself or Render properties?

ethanmoffat: I'm really not sure of a good solution for this. I'm not devoting a lot of thinking time to it because eolib-dotnet will remove the concept of "packet translators" from EndlessClient.

sorokya:
These issues are fixed in the new use_sdk branch

ethanmoffat:
Closing this PR; these issues are addressed by the migration to eolib-dotnet and confirmed fixed.

PR #355, opened 2024-06-01
Barber

AstralSkies:


EOLib.Localization/EOResourceID.cs
```diff
@@ -152,8 +152,13 @@ public enum EOResourceID
...
         DIALOG_SHOP_CRAFT_PUT_INGREDIENTS_TOGETHER = 159,
 
+        DIALOG_BARBER_CHANGE_MODAL = 160,
+        DIALOG_BARBER_CHANGE_HAIR_COLOR = 161,
+        DIALOG_BARBER_BUY_HAIRSTYLE = 162,
         DIALOG_WORD_CURRENT = 163,
-
+        DIALOG_BARBER_DO_YOU_WANT_TO_BUY_A_NEW_HAIRSTYLE = 164,
+        
+        DIALOG_TRADE_PLAYERS_NEED_TO_TRADE_ONE_ITEM = 165,
```
ethanmoffat, line 161: This new line is a duplicate of the resource below it and doesn't appear to be used in this PR? I think this line can be deleted

EOLib/Domain/Interact/Bank/BankDataRepository.cs
```diff
@@ -1,34 +1,47 @@
-﻿using AutomaticTypeMapper;
+﻿using System.Diagnostics;
+using AutomaticTypeMapper;
```
ethanmoffat, line 2: Changes in this file can be reverted in their entirety

EOLib.Localization/EOResourceID.cs
```diff
@@ -342,6 +347,8 @@ public enum EOResourceID
         STATUS_LABEL_IS_ONLINE_NOT_FOUND = 430,
         DIALOG_BANK_LOCKER_UPGRADE = 431,
         DIALOG_BANK_MORE_SPACE = 432,
-        STATUS_LABEL_LOCKER_SPACE_INCREASED = 433
+        STATUS_LABEL_LOCKER_SPACE_INCREASED = 433,
+
```
ethanmoffat, line 351: nit: delete extra whitespace lines

EOLib/Domain/Interact/Barber/BarberDataRepository.cs
```diff
@@ -0,0 +1,40 @@
...
+    [AutoMappedType(IsSingleton = true)]
+    public class BarberDataRepository : IBarberDataRepository, IBarberDataProvider
+    {
+        private int _sessionID;
+        public int SessionID
+        {
+            get => _sessionID;
+            set
+            {
+                _sessionID = value;
```
ethanmoffat, line 25: Make this an auto-property:

`public int SessionID { get; set; }`

EOLib/PacketHandlers/Avatar/AvatarAgreeHandler.cs
```diff
@@ -17,7 +17,7 @@ public class AvatarAgreeHandler : InGameOnlyPacketHandler
         protected readonly ICurrentMapStateRepository _currentMapStateRepository;
         protected readonly ICharacterRepository _characterRepository;
         protected readonly IEIFFileProvider _eifFileProvider;
-
+
```
ethanmoffat, line 20: Changes in this file can be reverted in their entirety

EOLib/PacketHandlers/Barber/BarberReplyHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+using EOLib.Domain.Interact.Barber;
+using EOLib.Domain.Character;
+using EOLib.Domain.Login;
+using EOLib.Domain.Map;
+using EOLib.Domain.Notifiers;
+
+namespace EOLib.PacketHandlers.Barber
+{
+    [AutoMappedType]
+    public class BarberReplyHandler : InGameOnlyPacketHandler
```
ethanmoffat, line 15: Rename type/file to `BarberAgreeHandler` to match the packet ID

EOLib/PacketHandlers/Shop/ShopTradeHandler.cs
```diff
@@ -109,4 +109,4 @@ public class ShopSellHandler : ShopTradeHandler
         {
         }
```
ethanmoffat, line 110: Changes in this file can be reverted in their entirety

EndlessClient.sln
```diff
@@ -13,7 +13,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EOLib", "EOLib\EOLib.csproj
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EOLib.Config", "EOLib.Config\EOLib.Config.csproj", "{B6AD3196-D307-4352-9979-9237CB08B102}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "EOLib.Graphics", "EOLib.Graphics\EOLib.Graphics.csproj", "{8C4BC247-89ED-4629-B34B-15A5AEE88E2D}"
```
ethanmoffat, line 16: Not sure what's happening here, can these changes be reverted?

EndlessClient/ControlSets/InitialControlSet.cs
```diff
@@ -159,6 +159,8 @@ private IXNALabel GetVersionInfoLabel()
         private IXNAPictureBox GetPersonPicture1()
         {
             var texture = _personSet1[_randomGen.Next(4)];
+
```
ethanmoffat, line 162: Changes to this file can be reverted in their entirety

EndlessClient/Dialogs/Actions/InGameDialogActions.cs
```diff
@@ -12,6 +12,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Diagnostics;
```
ethanmoffat, line 15: nit: remove unused using

EndlessClient/Dialogs/Actions/InGameDialogActions.cs
```diff
@@ -69,7 +71,9 @@ public class InGameDialogActions : IInGameDialogActions
                                    ILawDialogFactory lawDialogFactory,
                                    IHelpDialogFactory helpDialogFactory,
                                    ISfxPlayer sfxPlayer,
-                                   IStatusLabelSetter statusLabelSetter)
+                                   IStatusLabelSetter statusLabelSetter,
+                                   IBarberDialogFactory barberDialogFactory)
+
```
ethanmoffat, line 76: nit: remove whitespace

EndlessClient/Dialogs/Actions/NpcInteractionActions.cs
```diff
@@ -11,6 +11,7 @@
 using EOLib.Localization;
 using System;
 using XNAControls;
+using System.Diagnostics;
```
ethanmoffat, line 14: nit: remove unused using

EndlessClient/Dialogs/ActiveDialogRepository.cs
```diff
@@ -165,6 +171,8 @@ IReadOnlyList<Option<IXNADialog>> ActiveDialogs
                     JukeboxDialog.Map(Map),
                     InnkeeperDialog.Map(Map),
                     LawDialog.Map(Map),
+                    BarberDialog.Map(Map),
+                    BardDialog.Map(Map),
```
ethanmoffat, line 175: `BardDialog.Map` is already in this list

EndlessClient/Dialogs/BaseEODialog.cs
```diff
@@ -12,6 +12,8 @@ public abstract class BaseEODialog : XNADialog
 
         public INativeGraphicsManager GraphicsManager { get; }
 
+        public ListDialogItem.ListItemStyle ListItemType { get; set; }
```
ethanmoffat, line 15: This shouldn't be in `BaseEODialog`. Not every dialog has list items.

EndlessClient/Rendering/Character/CharacterAnimator.cs
```diff
@@ -17,6 +17,7 @@
 using System.Diagnostics;
 using System.Linq;
 
+
```
ethanmoffat, line 20: Changes to this file can be removed in their entirety

EndlessClient/Rendering/Sprites/SpriteSheet.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
+using System.Diagnostics;
```
ethanmoffat, line 3: Changes to this file can be removed in their entirety

EndlessClient/UIControls/CreateCharacterControl.cs
```diff
@@ -72,5 +72,22 @@ private static CharacterRenderProperties GetDefaultProperties()
         {
             return new CharacterRenderProperties.Builder { HairStyle = 1 }.ToImmutable();
         }
+
+        public void UpdateRenderProperties(int hairStyle, int hairColor, int race, int gender, int armorgraphic, int weapongraphic, int bootgraphic, int shieldgraphic)
```
ethanmoffat, line 76: Instead of adding an `UpdateRenderProperties` method, add a constructor overload (and factory function overload) that takes in the current character's render properties instead of using default properties. Then you can just construct the object directly in `BarberDialog` from the main character's render properties.

It looks like barber dialog doesn't need to arbitrarily set the render properties, and I'd prefer not to have a mutating method like this on the control's interface.

For clarify, I'm thinking like this:
```csharp
public CreateCharacterControl(ICharacterRendererFactory characterRendererFactory)
    : this(GetDefaultProperties(), characterRendererFactory) { }

public CreateCharacterControl(CharacterRenderProperties renderProperties, ICharacterRendererFactory factory)
    : base(Character.Default.WithRenderProperties(renderProperties), factory
{
    SetSize(99, 123);
    _lastPosition = Vector2.Zero;
}
```

EndlessClient/Dialogs/ListDialogItem.cs
```diff
@@ -18,6 +18,10 @@ public enum ListItemStyle
             Large
         }
 
+
+        public bool EnableBarberMode = false;
```
ethanmoffat, line 22: `ListDialogItem` should not have anything barber-specific. This is a general control for all list dialog items. Also, convention is to avoid public fields and use properties instead.

If you want to override the width of the highlight, add a property that sets it, similar to `OffsetX` and `OffsetY`.

`OffsetX` already performs the function you're doing in the `Draw` override for getting `rectX`.

EndlessClient/UIControls/BlinkingLabel.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
```
ethanmoffat, line 2: I don't see the value in these changes - is there something wrong with `BlinkingLabel`? Strongly prefer to revert this file in its entirety.

Usually a crash in `base.Update()` shows up in `EndlessGame` due to collections being modified. This is *usually* caused by trying to update the collection on a background thread (collection was modified, enumeration may not continue).

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        public override void Initialize()
+        {
+            base.Initialize();
+            _characterControl.Initialize();
+            UpdateListItems(_characterRepository.MainCharacter.RenderProperties);
+
+        }
+        public override void Update(GameTime gameTime)
```
ethanmoffat, line 94: This override is unneeded

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,223 @@
...
+            var message = $"{_localizedStringFinder.GetString(EOResourceID.DIALOG_BARBER_DO_YOU_WANT_TO_BUY_A_NEW_HAIRSTYLE)}, {cost} {_eifFileProvider.EIFFile[1].Name}";
+            var title = _localizedStringFinder.GetString(EOResourceID.DIALOG_BARBER_BUY_HAIRSTYLE);
+            var msgBox = _messageBoxFactory.CreateMessageBox(message, title, EODialogButtons.OkCancel);
+
+            msgBox.DialogClosing += (_, e) =>
+            {
+                if (e.Result == XNADialogResult.OK)
+                {
+                    _barberActions.Purchase(hairStyle, hairColor);
+                    foreach (var notifier in _soundNotifiers)
```
ethanmoffat, line 208: Notifiers are generally used for signalling from `EOLib` up to `EndlessClient` that something happened (because I wanted to re-invent event handlers, I guess, I regret the design decision but I digress)

You can just inject `ISfxPlayer` and play `SoundEffectID.BuySell`. See jukebox dialog.

EndlessClient/Dialogs/Factories/HelpDialogFactory.cs
```diff
@@ -53,13 +53,11 @@ private string[] GetMessages()
         {
             return new[]
             {
-                _localizedStringFinder.GetString(EOResourceID.ENDLESS_HELP_SUMMARY_1),
```
ethanmoffat, line 56: Why were these changes deleted? Can this file be reverted?

EndlessClient/Dialogs/ScrollingListDialog.cs
```diff
@@ -74,7 +74,7 @@ public enum DialogType
         BankAccountDialog,
 
         Jukebox,
-        Barber,
```
ethanmoffat, line 77: If `Barber` isn't going to be a `ScrollingListDialog` then this is fine, remove all references to `Barber` from this file

EndlessClient/Dialogs/ScrollingListDialog.cs
```diff
@@ -121,7 +121,7 @@ public int ItemsToShow
             }
         }
 
-        public ListDialogItem.ListItemStyle ListItemType
+        public new ListDialogItem.ListItemStyle ListItemType
```
ethanmoffat, line 124: This should be reverted along with removing `ListItemType` from `BaseEoDialog`

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        private void Hair(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e) => UpdateHairStyle(1);
+        private void HairColor(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e) => UpdateHairColor(1);
+
+        private void UpdateHairStyle(int offset)
+        {
+            var currentProperties = _characterControl.RenderProperties;
+            var newHairStyle = (currentProperties.HairStyle + offset) % 21;
+            _characterControl.UpdateRenderProperties(newHairStyle, currentProperties.HairColor,
```
ethanmoffat, line 160: `CreateCharacterControl.NextHairStyle` already does this.

Note that the base class `CharacterControl` has a publicly-visible `RenderProperties` property that you can access to get the current hairstyle for setting the subtext

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,223 @@
...
+                                                     currentProperties.ArmorGraphic, currentProperties.WeaponGraphic,
+                                                     currentProperties.BootsGraphic, currentProperties.ShieldGraphic);
+            _changeHairItem.SubText = GetCurrentHairStyleText(newHairStyle);
+        }
+
+        private void UpdateHairColor(int offset)
+        {
+            var currentProperties = _characterControl.RenderProperties;
+            var newHairColor = (currentProperties.HairColor + offset) % _hairColorNames.Length;
+            _characterControl.UpdateRenderProperties(currentProperties.HairStyle, newHairColor,
```
ethanmoffat, line 171: `CreateCharacterControl.NextHairColor` already does this.

Note that the base class `CharacterControl` has a publicly-visible `RenderProperties` property that you can access to get the current hair color for setting the subtext

ethanmoffat (changes requested):

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,223 @@
...
+            _characterControl.SetParentControl(this);
+            _characterControl.UpdateRenderProperties(currentProperties.HairStyle, currentProperties.HairColor,
+                                                     currentProperties.Race, currentProperties.Gender,
+                                                     currentProperties.ArmorGraphic, currentProperties.WeaponGraphic,
+                                                     currentProperties.BootsGraphic, currentProperties.ShieldGraphic);
+        }
+
+        private void InitializeDialogItems(IEODialogButtonService dialogButtonService)
+        {
+            var cancel = CreateButton(dialogButtonService, new Vector2(215, 150), SmallButton.Cancel);
```
ethanmoffat, line 81: I think y coordinate needs to be +1 here
image

EndlessClient/UIControls/BlinkingLabel.cs
AstralSkies: image

I get this issue if i revert or I can do the below instead

image

EndlessClient/Dialogs/Factories/HelpDialogFactory.cs
AstralSkies: leftovers on getting used to the repo haha. sorry !

EndlessClient/Dialogs/ListDialogItem.cs
```diff
@@ -69,6 +66,13 @@ public int OffsetY
             }
         }
 
+        public int HighlightWidthOverride
```
ethanmoffat, line 69: 1. Can this be an auto property? Trying to avoid the property-with-backing-field paradigm unless the getter/setter have logic associated with the property and it is unavoidable.
2. Can you make this an `Optional<int>` or `int?` instead of using the magic value `-1` to indicate that it is unset?

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            _characterControl.NextHairStyle();
+            _changeHairItem.SubText = GetCurrentHairStyleText(_characterControl.RenderProperties.HairStyle);
+        }
+
+        private void HairColor(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e)
+        {
+            UpdateHairColor();
+        }
+        private void UpdateHighlightWidth(ListDialogItem item)
```
ethanmoffat, line 163: It looks like this method is unused

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            item.HighlightWidthOverride = 175;
+        }
+        private void UpdateHairColor()
```
ethanmoffat, line 167: You can inline this method, it is only called once

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            var currentProperties = _characterControl.RenderProperties;
+            _characterControl.NextHairColor();
+            _changeHairColor.SubText = GetCurrentHairColorText(_characterControl.RenderProperties.HairColor);
+        }
+
+        private void BuyHairStyleOrColor(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e) => PurchaseHairStyleOrColor(_characterControl.RenderProperties.HairStyle, _characterControl.RenderProperties.HairColor);
+
+        private void PurchaseHairStyleOrColor(int hairStyle, int hairColor)
```
ethanmoffat, line 176: Also prefer to inline this method. I don't think having the event handles hook into a different method really adds much clarity.

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+        private string GetCurrentHairColorText(int hairColor) => $"{_localizedStringFinder.GetString(EOResourceID.DIALOG_WORD_CURRENT)}: {_hairColorNames[hairColor % _hairColorNames.Length]}";
+        private string GetCostText() => $"{_localizedStringFinder.GetString(EOResourceID.DIALOG_WORD_CURRENT)}: {CalculateCost()} gold";
+
+        private int CalculateCost()
+        {
+            var level = (int)_characterRepository.MainCharacter.Stats[CharacterStat.Level];
+            return 200 + Math.Max(level - 1, 0) * 200;
+        }
+
+        private void Hair(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e)
```
ethanmoffat, line 153: Rename the event handlers something like `ChangeHairStyle_Click`, `ChangeHairColor_Click`, `BuyHair_Click` so it's clearer from the names that they're vent handlers.

EndlessClient/Dialogs/ListDialogItem.cs
```diff
@@ -20,6 +21,8 @@ public enum ListItemStyle
 
         private int _index;
         private int _xOffset, _yOffset;
+        private int _highlightWidthOverride = -1;
+        private int _highlightXOffset = 0;
```
ethanmoffat, line 25: Delete this unused field

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        public override void Initialize()
+        {
+            base.Initialize();
+            _characterControl.Initialize();
+            UpdateListItems(_characterRepository.MainCharacter.RenderProperties);
+        }
+
+        public void Update(GameTime gameTime)
```
ethanmoffat, line 90: Remove this unused method (this also doesn't have `override` so it could cause issues too)

EndlessClient/Dialogs/BarberDialog.cs
ethanmoffat: This doesn't appear to be resolved with the latest changes (the thick black line at the bottom of the button is incorrect). Change to `new Vector2(215, 151)`

ethanmoffat (changes requested):

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,209 @@
...
+            ApplyWidthOverride(_changeBuyHairStyleOrColor);
+
+            ApplyHighlightOffsetXOverride(_changeHairItem);
+            ApplyHighlightOffsetXOverride(_changeHairColor);
+            ApplyHighlightOffsetXOverride(_changeBuyHairStyleOrColor);
+        }
+
+        private void ApplyWidthOverride(ListDialogItem item)
+        {
+            item.WidthOverride = AdjustedWidth;
```
ethanmoffat, line 108: I'm thinking that maybe `ListDialogItem` doesn't need to change. The `WidthOverride` value is set here, but this property isn't actually used anywhere (that I can see) except in the `SetItemSize` call below. `SetItemSize` is a wrapper around the protected method `SetSize`, which is just a convenience function for updating the `DrawArea` without changing the coordinates from within the control.

1. Back out changes to `ListDialogItem`(same as whitespace process, but for `EndlessClient/Dialogs/ListDialogItem.cs`).
2. Change the implementation of `ApplyWidthOverride` here to
```cs
item.DrawArea = new Rectangle(item.DrawArea.X, item.DrawArea.Y, AdjustedWidth, item.DrawArea.Height);
```

For style points, also do the following:
1. Mark this method as `static` (actually, double-check the other helper methods that you added and make sure they're `static` if they don't access instance members)
2. Pass in the width as a parameter to this method. This has the benefit of making the method call in `CreateListItems` a bit more readable, rather than implicitly relying on a constant defined elsewhere, it's now explicitly passed in.

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,209 @@
...
+            item.SetItemSize(item.WidthOverride.GetValueOrDefault(232), item.Style == ListDialogItem.ListItemStyle.Large ? 36 : 13); 
+        }
+
+        private void ApplyHighlightOffsetXOverride(ListDialogItem item)
+        {
+            item.OffsetX = AdjustedHighlightXOffset;
```
ethanmoffat, line 114: I appreciate that you're trying to make the code more readable by splitting out a bunch of helper methods. I don't know if these are really adding a lot of clarity though, especially for these one-line things. IMO it's probably better to just initialize the list items in-line so readers (me) don't have to decode the call stack and trace through what's actually happening between all these calls.

I'll leave it up to your discretion on this one (comment here with your decision so I know which direction you think is better), but I think just constructing the list items in one spot with a property initializer list would make this a lot easier to parse through.

EndlessClient/UIControls/CreateCharacterControl.cs
```diff
@@ -68,6 +74,13 @@ public void NextHairColor()
             RenderProperties = RenderProperties.WithHairColor((RenderProperties.HairColor + 1) % 10);
         }
 
+        public void NextDirection()
```
ethanmoffat, line 77: Is this method invoked anywhere? My limited "Ctrl+F" PR review search ability isn't turning up anything. If it's not used, please delete

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        private void ChangeHairStyle_Click(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e)
+        {
+            _characterControl.NextHairStyle();
+            _changeHairItem.SubText = GetCurrentHairStyleText(_characterControl.RenderProperties.HairStyle);
+        }
+
+        private void ChangeHairColor_Click(object sender, MonoGame.Extended.Input.InputListeners.MouseEventArgs e)
+        {
+            var currentProperties = _characterControl.RenderProperties;
```
ethanmoffat, line 170: Unused variable here.

Side note: check out the compiler output and make sure there are no warnings.

EndlessClient/UIControls/BlinkingLabel.cs
ethanmoffat: Resolved with `isUiControl` parameter when creating character renderers (to be fixed more elegantly later)

ethanmoffat (changes requested):

EndlessClient/Dialogs/BarberDialog.cs
ethanmoffat: I have no idea why the compiler isn't flagging this; maybe it's because it's just a reference? Weird.

EndlessClient/Dialogs/BarberDialog.cs
AstralSkies: "every abstraction has its cost (readability in this case)". I'll go inline.

EndlessClient/Dialogs/BarberDialog.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            _characterControl.SetParentControl(this);
+        }
+
+        private void InitializeDialogItems(IEODialogButtonService dialogButtonService)
+        {
+            var cancel = CreateButton(dialogButtonService, new Vector2(215, 151), SmallButton.Cancel);
+            cancel.OnClick += (_, _) => Close(XNADialogResult.Cancel);
+
+            _changeHairItem = new ListDialogItem(this, ListDialogItem.ListItemStyle.Large, 0)
```
ethanmoffat, line 81: Ok, last thing (probably):

- `OffsetY` is `25` by default for large style items. This does not need to be set here.
- Pass in `0`, `1`, and `2` as the list index for each subsequent list item (third constructor parameter, currently all zero). It automatically does the math for offsetting the list items (36*index for large style).
- Include `.DrawArea = ` and `.OffsetX = ` in the initializer lists

AstralSkies: You will end up with the incorrect data if you add include .DrawArea in the initializer lists as there's no method of pre-knowing the math's for the draw size.

ethanmoffat (approved):

PR #357, opened 2024-06-06
Show admin hide effect when player hides/unhides

### LINKED ISSUE #351: Admins hiding does not play effect
There is a "bubble" effect for admins hiding and a sound effect. The effect metadata is already set up in `EffectMetadataProvider`, but a hide (or unhide) action by an admin does not trigger it.

sorokya:
Closes #351 

@ethanmoffat is there a better way to reference the effect ID here besides a magic number?

### DIFF
--- a/EndlessClient/Rendering/Character/CharacterAnimationActions.cs
+++ b/EndlessClient/Rendering/Character/CharacterAnimationActions.cs
@@ -335,6 +335,13 @@ namespace EndlessClient.Rendering.Character
                     .RenderEffect(location, effectId);
             }
         }
+        public void NotifyAdminHideEffect(int playerId)
+        {
+            if (playerId == _characterRepository.MainCharacter.ID)
+                _characterRendererProvider.MainCharacterRenderer.MatchSome(cr => cr.PlayEffect((int)HardCodedEffect.AdminHide));
+            else if (_characterRendererProvider.CharacterRenderers.ContainsKey(playerId))
+                _characterRendererProvider.CharacterRenderers[playerId].PlayEffect((int)HardCodedEffect.AdminHide);
+        }
 
         private void ShowWaterSplashiesIfNeeded(CharacterActionState action, int characterID)
         {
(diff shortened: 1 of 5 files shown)


ethanmoffat:
> Closes #351
> 
> @ethanmoffat is there a better way to reference the effect ID here besides a magic number?

There's `HardCodedEffect` that I have for things like the potion effects. Also includes `WarpLeave` and `WarpArrive`. https://github.com/ethanmoffat/EndlessClient/blob/d5f218181ebfa1a97a9afa7c2df801d9155f7573/EndlessClient/Rendering/Effects/HardCodedEffect.cs#L3

That's in rendering code though (not library code). I'd suggest changing the approach you've taken to add a method call `ShowAdminHideEffect(int playerId)` to the `IEffectNotifier` interface. This is implemented in `CharacterAnimationActions` so you could basically copy/paste the WarpEnter or WarpLeave implementation for `ShowAdminHideEffect`, just use a different `HardCodedEffect`. `NotifyPotionEffect` is currently really only used for potions, even though it could be used for any effect I'd prefer to keep it limited to just playing a potion effect.

sorokya:
@ethanmoffat lol I didn't even realize I was using `NotifyPotionEffect` at the time. Updated code to use new effect notifier and hard coded effect id.

EndlessClient/Rendering/Character/CharacterAnimationActions.cs
```diff
@@ -415,6 +415,14 @@ private bool IsSteppingStone(CharacterRenderProperties renderProps)
                 || _currentMapProvider.CurrentMap.Tiles[renderProps.GetDestinationY(), renderProps.GetDestinationX()] == TileSpec.Jump;
         }
 
+        public void NotifyAdminHideEffect(int playerId)
```
ethanmoffat, line 418: Move this method up to where the rest of the public interface is defined, general style is to keep public/private methods grouped together in a file

ethanmoffat (changes requested):
Actual change looks good, just move the method up to where the rest of the public methods are defined

EndlessClient/Rendering/Character/CharacterAnimationActions.cs
sorokya: @ethanmoffat done. Sorry it took a month :P

ethanmoffat (approved):

PR #358, opened 2024-06-06
Chat fixes

### LINKED ISSUE #352: Chat bugs and inconsistencies
The following chat-related bugs exist:
- Public and Group chat always shows speech bubble over the main character
- Admin chat does not play sfx when you send a message, only when you receive one
- Group chat does not play sfx when you are on a different map as the sender
- PM tab [x] button doesn't work in resizable display mode
- PM recipient not found doesn't play the SFX

sorokya:
Closes #352 

> PM recipient not found doesn't play the SFX

I couldn't replicate this. The SFX played for me

> Public and Group chat always shows speech bubble over the main character

This was a bug where we were always overwriting the render character with the main character. Now we always find one in `_characterRendererProvider.CharacterRenderers` with the matching player id

> Admin chat does not play sfx when you send a message, only when you receive one

Added sound effect to chat controller when sending message

> Group chat does not play sfx when you are on a different map as the sender

We were only looking at party members in the local map repository instead of the party repository before. Now it shows messages from members no matter where they are in game.

> PM tab [x] button doesn't work in resizable display mode

I believe this was a render order bug? Seemed like some kinda z-index kind of thing. Adding the `Tab` clickable area before the close buttons fixed it. I also adjusted the dimensions a bit to match the gfx.

### DIFF
--- a/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
+++ b/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
@@ -7,9 +7,6 @@ using EOLib.Domain.Party;
 using EOLib.Net.Handlers;
 using Moffat.EndlessOnline.SDK.Protocol.Net;
 using Moffat.EndlessOnline.SDK.Protocol.Net.Server;
-using System.Collections.Generic;
-using System.Linq;
-using Optional;
 using Optional.Collections;
 
 namespace EOLib.PacketHandlers.Chat
--- a/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
+++ b/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
@@ -40,7 +40,7 @@ namespace EOLib.PacketHandlers.Chat
 
         public override bool HandlePacket(TalkOpenServerPacket packet)
         {
-            _partyDataProvider.Members.FirstOrNone(member => member.CharacterID == packet.PlayerId)
+            _partyDataProvider.Members.SingleOrNone(member => member.CharacterID == packet.PlayerId)
                 .MatchSome(member =>
                 {
                     var localChatData = new ChatData(ChatTab.Local, member.Name, packet.Message, ChatIcon.PlayerPartyDark, ChatColor.PM);
--- a/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
+++ b/EOLib/PacketHandlers/Chat/GroupChatHandler.cs
@@ -9,6 +9,8 @@ using Moffat.EndlessOnline.SDK.Protocol.Net;
 using Moffat.EndlessOnline.SDK.Protocol.Net.Server;
 using System.Collections.Generic;
 using System.Linq;
+using Optional;
+using Optional.Collections;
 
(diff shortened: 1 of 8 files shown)


ethanmoffat:
> Closes #352
> 
> > PM recipient not found doesn't play the SFX
> 
> I couldn't replicate this. The SFX played for me

I was having a bunch of issues with PMs, so it might work inconsistently. The chat code was some of the first code I rewrote and it has not aged well.

EOLib/PacketHandlers/Chat/GroupChatHandler.cs
```diff
@@ -1,55 +1,59 @@
...
+            _partyDataProvider = partyDataProvider;
             _otherCharacterEventNotifiers = otherCharacterEventNotifiers;
             _chatEventNotifiers = chatEventNotifiers;
         }
 
         public override bool HandlePacket(TalkOpenServerPacket packet)
         {
-            return Handle(packet, packet.PlayerId);
-        }
+            var member = _partyDataProvider.Members.FirstOrDefault(member => member.CharacterID == packet.PlayerId);
```
ethanmoffat, line 40: I've been trying to use optional pattern - `OptionalFirst(x => x.CharacterID == packet.PlayerId).MatchSome(x => ...)` - instead of linq's `XOrDefault` methods with comparison to `null`. More functional and all.

ethanmoffat:
Can you add some details to the PR description about *why* the fixes work? You have the "what" in the commit history, but I'd like a clear line between "I changed this" and "this is the effect"

EOLib/PacketHandlers/Chat/GroupChatHandler.cs
sorokya: Can you share an example of what you would do here? I'm not familiar with that I guess

sorokya:
@ethanmoffat PR description updated and can you show an example of how you'd replace the `FirstOrDefault` usage?

EOLib/PacketHandlers/Chat/GroupChatHandler.cs
ethanmoffat: It's used all over the code base - if you search for `OptionalFirst` solution-wide you'll see the pattern.

There isn't much more to it than the example I gave. It's exactly like `SingleOrDefault` in that you pass a predicate to the method that provides a match. The result is an `Option<T>`, which is sort of like `OneOf` except it's either "Some" or "None". The chained call `MatchSome(x => ...)` does some action when there is some value.

It's used as a way to avoid the possibility of `NullReferenceException` being thrown due to a null value. It's a clear way to mark intent and force the caller to handle the return value regardless of whether there is one.

sorokya: @MrDanOak has fixed this now using `SingleOrNone` and `MatchSome` :)

ethanmoffat: Looks good - if you guys sanity check this one more time to make sure the bugs in 352 are all no longer reproducible then I'm cool with merging this

ethanmoffat (approved):

PR #359, opened 2024-06-06
Send WALK_SPEC for ghost walking

### LINKED ISSUE #349: Send WALK_SPEC when ghosting players
When ghosting another player, the `WALK_SPEC` packet is used in place of the `WALK_PLAYER` packet. The contents are identical.

Oversight in initial ghosting implementation.

sorokya:
Closes #349 

This is disgusting and I'm ashamed. Please tell me how to do it better lmao

### DIFF
--- a/EndlessClient/Controllers/ArrowKeyController.cs
+++ b/EndlessClient/Controllers/ArrowKeyController.cs
@@ -128,8 +128,10 @@ namespace EndlessClient.Controllers
                     _unwalkableTileActionsHandler.HandleUnwalkableTileActions(unwalkableActions, cellState);
                     break;
                 case WalkValidationResult.GhostComplete:
+                    _characterAnimationActions.StartWalking(Option.None<MapCoordinate>(), true);
+                    break;
                 case WalkValidationResult.Walkable:
-                    _characterAnimationActions.StartWalking(Option.None<MapCoordinate>());
+                    _characterAnimationActions.StartWalking(Option.None<MapCoordinate>(), false);
                     break;
             }
         }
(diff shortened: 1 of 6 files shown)


ethanmoffat:
> Closes #349
> 
> This is disgusting and I'm ashamed. Please tell me how to do it better lmao

Honestly probably not much different than I would have done it. You're kinda stuck doing it this way because `CharacterAnimator` is spaghetti. My only suggestion would be to add explicit methods for `WalkGhost` but that doesn't really make a ton of sense.

EndlessClient/Rendering/Character/CharacterAnimationActions.cs
```diff
@@ -422,7 +422,7 @@ public interface ICharacterAnimationActions
     {
         void Face(EODirection direction);
 
-        void StartWalking(Option<MapCoordinate> targetCoordinate);
+        void StartWalking(Option<MapCoordinate> targetCoordinate, bool ghosted);
```
ethanmoffat, line 425: I think a default value of `=false` makes sense both here and in `CharacterActions`.

EndlessClient/Controllers/ArrowKeyController.cs
```diff
@@ -128,6 +128,8 @@ private void AttemptToStartWalking()
                     _unwalkableTileActionsHandler.HandleUnwalkableTileActions(unwalkableActions, cellState);
                     break;
                 case WalkValidationResult.GhostComplete:
+                    _characterAnimationActions.StartWalking(Option.None<MapCoordinate>(), true);
```
ethanmoffat, line 131: Can you add the explicit parameter notation here, so it's more obvious what the bool is? `, ghosted: true` instead of just `, true`

ethanmoffat (changes requested):
You'll have to add the `ghosted: false` parameter to the call in EOBot, so the solution keeps building

EndlessClient/Controllers/ArrowKeyController.cs
sorokya: @ethanmoffat fixed!

ethanmoffat (approved):

PR #362, opened 2024-06-15
Reboot handler

sorokya:
Adds packet handler for Message_Close and event notifier/subscriber.

Also tried making disconnect detection better but it doesn't really work for some reason.

Found this lengthy thread of people discussing the problem
https://stackoverflow.com/questions/722240/instantly-detect-client-disconnection-from-server-socket

In reoserv I do this
```rust
match self.socket.try_read(&mut buf) {
	Ok(0) => { // <- This means the client disconnected
		return Some(Err(std::io::Error::new(
			std::io::ErrorKind::BrokenPipe,
			"Connection closed",
		)));
	}
	Ok(_) => {}
	Err(_) => {
		return None;
	}
}
```

I tried doing something similar in EndlessClient but it was getting false positives :\
```c#
!(_socket.Poll(1, SelectMode.SelectRead) && _socket.Available == 0);
```

edit: actually that method would probably work but the polling timeout needs to be higher and we probably wouldn't want that in the main thread.. idk something to think about

Anyway.. this handler/sfx/server message is wired up now :)

### DIFF
--- a/EOLib/Domain/Notifiers/IServerRebootNotifier.cs
+++ b/EOLib/Domain/Notifiers/IServerRebootNotifier.cs
@@ -1,0 +1,15 @@
+﻿using AutomaticTypeMapper;
+
+namespace EOLib.Domain.Notifiers
+{
+    public interface IServerRebootNotifier
+    {
+        void NotifyServerReboot();
+    }
+
+    [AutoMappedType]
+    public class NoOpServerRebootNotifier : IServerRebootNotifier
+    {
+        public void NotifyServerReboot() { }
+    }
+}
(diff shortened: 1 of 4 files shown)


ethanmoffat:
The lack of immediate disconnect detection is pretty annoying to me. Generally if the server drops the connection, you won't see the effect until you try and do something that performs a network action client-side. It'd be great if it was handled better.

You can see my attempt at it here: https://github.com/ethanmoffat/EndlessClient/blob/d5f218181ebfa1a97a9afa7c2df801d9155f7573/EndlessClient/Network/PacketHandlerGameComponent.cs#L36

I think I used to have the `SelectRead`/`Available` pair of calls in `NetworkClient` but was similarly having issues with it.

ethanmoffat (approved):

PR #363, opened 2024-07-23
Guild creation request

sorokya:
Implements guild creation request packet handler and shows dialog


EOLib/Domain/Interact/Guild/GuildCreationRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace EOLib.Domain.Interact.Guild
+{
+    public class GuildCreationRequest
```
ethanmoffat, line 3: Look at the other record types (decorated with `[Record]` in EOLib.Domain) - this should follow the immutable pattern

EndlessClient/Subscribers/GuildEventSubscriber.cs
```diff
@@ -0,0 +1,54 @@
...
+                _localizedStringFinder.GetString(EOResourceID.GUILD_DO_YOU_ACCEPT)
+                , caption: _localizedStringFinder.GetString(DialogResourceID.GUILD_INVITATION), whichButtons: Dialogs.EODialogButtons.OkCancel,
+                style: Dialogs.EOMessageBoxStyle.LargeDialogSmallHeader);
+
+            dlg.DialogClosing += (_, e) =>
+            {
+                if (e.Result == XNAControls.XNADialogResult.OK)
+                {
+                    var packet = new GuildAcceptClientPacket();
+                    packet.InviterPlayerId = request.CreatorPlayerID;
```
ethanmoffat, line 46: Minor nit: use the member initialization syntax here

```cs
var packet = new GuildAcceptClientPacket { InviterPlayerId = request.CreatorPlayerID };
```

EndlessClient/Subscribers/GuildEventSubscriber.cs
```diff
@@ -0,0 +1,54 @@
...
+            _packetSendService = packetSendService;
+            _sfxPlayer = sfxPlayer;
+        }
+
+        public void NotifyGuildCreationRequest(GuildCreationRequest request)
+        {
+            _sfxPlayer.PlaySfx(SoundEffectID.ServerMessage);
+
+            var dlg = _messageBoxFactory.CreateMessageBox(
+                $"{request.GuildIdentity}" + " " +
```
ethanmoffat, line 33: Minor nit: formatting. Put spaces as the first part of the line, indent subsequent lines for the string, comma does not start the `caption` line

```cs
$"{request.GuildIdentity}"
    + " " + _localizedStringFinder.GetString(...)
    + " " + ...
    + " " + _localizedStringFinder.GetString(EOResourceID.GUILD_DO_YOU_ACCEPT),
caption: ...
```

EndlessClient/Subscribers/GuildEventSubscriber.cs
```diff
@@ -0,0 +1,54 @@
...
+                _localizedStringFinder.GetString(DialogResourceID.GUILD_INVITES_YOU_TO_JOIN) + " " +
+                _localizedStringFinder.GetString(EOResourceID.GUILD_JOINING_A_GUILD_IS_FREE) + " " +
+                _localizedStringFinder.GetString(EOResourceID.GUILD_PLEASE_CONSIDER_CAREFULLY) + " " +
+                _localizedStringFinder.GetString(EOResourceID.GUILD_DO_YOU_ACCEPT)
+                , caption: _localizedStringFinder.GetString(DialogResourceID.GUILD_INVITATION), whichButtons: Dialogs.EODialogButtons.OkCancel,
```
ethanmoffat, line 38: I'm really torn on this. Every other `DialogResourceID` specifies the first resource and uses both as a pair. This is kind of a special case because there's so much extra text.

If you take a look at the overloads for `EOMessageBoxFactory.CreateMessageBox` you'll see that there are some that allow putting extra text in there.

If you wouldn't mind, I'd appreciate if you could investigate how ugly it would be to use/add one that incorporates the additional text that this dialog box requires. If you're spending more than like 10 minutes on it I'm fine with accepting this as it is, but I'd like to avoid introducing the second resource for the pair of `DialogResourceID` values if at all possible.

ethanmoffat (changes requested):

EOLib/Domain/Interact/Guild/GuildCreationRequest.cs
sorokya: Yeah tbh I'm thinking of just ditching this and sending the values as arguments. That sound good?

ethanmoffat: I like the idea of a structure that encapsulates the parameters, but either way is fine if you want to go the other route. Separate parameters is also how the other notifiers do it so it's nice for consistency.

EndlessClient/Subscribers/GuildEventSubscriber.cs
sorokya: Yeah the one that accepts extra data isn't really going to work without modification since it prepends the guild identity before the second dialog resource string :\

ethanmoffat (approved):

EndlessClient/Subscribers/GuildEventSubscriber.cs
ethanmoffat: Thanks for looking into it - that's kind of what I figured

PR #364, opened 2024-07-24
Guild reply join request

sorokya:
Implements reply handler for join request


ethanmoffat:
This PR appears dependent on #363 so I'll wait to review until that one is merged. I want to ensure we're keeping a linear file history for some of the files that are being introduced in both PRs.

sorokya:
Yeah I should have marked this is a draft PR oops

ethanmoffat (approved):

PR #365, opened 2024-07-25
Add management menu

sorokya:
Adds guild management menu and shows "You are not in a guild" alert if a player not in a guild tries to interact with any of the buttons.


EOLib/Domain/Character/CharacterRepository.cs
```diff
@@ -22,5 +24,13 @@ public class CharacterRepository : ICharacterRepository, ICharacterProvider
         public bool HasAvatar { get; set; } = true;
 
         public Character MainCharacter { get; set; }
+
+        public bool InGuild
+        {
+            get
+            {
+                return MainCharacter.GuildTag != "   ";
```
ethanmoffat, line 32: I think this property makes more sense as an auto property on `Character` rather than as part of the repository. For things like this use syntax `public bool InGuild => GuildTag != "   ";`

Also - is this the confirmed behavior of the official client?

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -136,7 +150,100 @@ private void SetState(GuildDialogState state)
...
+
+                        AddItemToList(disbandItem, sortList: false);
+                    }
+                    break;
+
             }
+
+        }
+
+        private void CheckAndChangeState(GuildDialogState state)
```
ethanmoffat, line 237: Rename to something clearer: e.g. `SetStateIfGuildMember` (feel free to choose a better name if you think of one)

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -136,7 +150,100 @@ private void SetState(GuildDialogState state)
                         AddItemToList(bankAccountItem, sortList: false);
                     }
                     break;
+
+                case GuildDialogState.Management:
+                    {
+                        ListItemType = ListDialogItem.ListItemStyle.Large;
+                        Buttons = ScrollingListDialogButtons.BackCancel;
+
+                        BackAction += (_, _) => SetState(GuildDialogState.Initial);
```
ethanmoffat, line 159: This is going to add the lambda to the `BackAction` event as an additional handler *every* time we change to management state. The impact should be seen if you click management->Back->Management->Back; setting a breakpoint should case it to hit `SetState` twice (although the second will be a no-op since current state == new state).

BackAction should only be set up once in the constructor and should call `SetState` conditionally to the expected prior state based on the current state. You could also push the state onto a stack and pop it off in `BackAction` unconditionally so you don't need to do any additional processing.

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -136,7 +150,100 @@ private void SetState(GuildDialogState state)
...
+                        manageRankingItem.RightClick += (_, _) => CheckAndChangeState(GuildDialogState.ManageRankings);
+
+                        AddItemToList(manageRankingItem, sortList: false);
+
+                        var assignRankItem = new ListDialogItem(this, ListDialogItem.ListItemStyle.Large, 0)
+                        {
+                            ShowIconBackGround = false,
+                            IconGraphic = _dialogIconService.IconSheet,
+                            IconGraphicSource = _dialogIconService.GetDialogIconSource(DialogIcon.GuildRanking),
+                            PrimaryText = _localizedStringFinder.GetString(EOResourceID.GUILD_RANKING),
```
ethanmoffat, line 194: Is this supposed to be the same text as the above item? (see line 180)

Remember I don't have friends so I'm totally unfamiliar with how the guild dialog is supposed to work 🤣

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -31,6 +39,8 @@ private enum GuildDialogState
         private readonly IContentProvider _contentProvider;
         private readonly ICurrentMapStateProvider _currentMapStateProvider;
         private readonly IENFFileProvider _enfFileProvider;
```
ethanmoffat, line 41: I don't see where `_enfFileProvider` is being used - can you do a quick check and remove any dependencies that aren't in use in the dialog? They can always be added as needed later

ethanmoffat (changes requested):

EOLib/Domain/Character/CharacterRepository.cs
sorokya: Can we put it on Character here? I thought it was part of eolib. It's not confirmed no just an assumption

EndlessClient/Dialogs/GuildDialog.cs
sorokya: BackAction only called once now

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Yeah it does use the same "Ranking" text :P

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Removed unused dependencies

EOLib/Domain/Character/CharacterRepository.cs
ethanmoffat: > Can we put it on Character here? I thought it was part of eolib.

I don't think I understand why this being part of eolib precludes adding the property there. Remember because of my dumbassery that `EOLib` in the context of dotnet is the shared library for EndlessClient where `Moffat.EndlessOnline.SDK` is what contains the generated code structures.

From a design perspective a character provider/repository should really only have a character in it. The repository isn't what owns the concept of a guild tag, the character is. If the character was in an unmodifiable library I'd still argue for an extension method on Character rather than putting a property on the repository.

Regarding `HasAvatar` - I genuinely don't remember why I did it this way but I don't like it 🤣 that should be refactored.

> It's not confirmed just an assumption

I'm assuming there's no better way to identify whether a player is part of a guild, then? Would the `GuildTag` ever be empty (`""` or `string.Empty`) or unset (`null`)?

sorokya: The server always sends 3 bytes no matter what. When not in a guild it's `   `

sorokya: I thought Character was part of the SDK for some reason. It's defined on `Character` now lmao

ethanmoffat (approved):

PR #366, opened 2024-07-25
Modify guild description

sorokya:
TODO: Rebase after #365 is merged. Merge #365 first


ethanmoffat:
Like the other PR, going to wait until #365 is in to review this one

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -63,11 +78,19 @@ private enum GuildDialogState
...
+                _guildDescription = _guildSessionProvider.GuildDescription;
+                SetState(GuildDialogState.Modify);
+            }
 
+            base.OnUpdateControl(gameTime);
+        }
+
+        private void SetState(GuildDialogState state)
+        {
             _state = state;
```
ethanmoffat, line 94: I think you unintentionally removed the `if (_state == state)` early return check here. Definitely don't want/need to be doing all that work if the state is already set correctly.

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -29,6 +28,11 @@ private enum GuildDialogState
         private readonly ILocalizedStringFinder _localizedStringFinder;
         private readonly ICharacterProvider _characterProvider;
         private readonly IEOMessageBoxFactory _messageBoxFactory;
+        private readonly IGuildSessionProvider _guildSessionProvider;
+        private readonly IGuildActions _guildActions;
+        private readonly ITextInputDialogFactory _textInputDialogFactory;
+
+        private string _guildDescription { get; set; }
```
ethanmoffat, line 35: This should be a private field, not a property

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -63,11 +78,19 @@ private enum GuildDialogState
...
 
-        private void SetState(GuildDialogState state)
+        protected override void OnUpdateControl(GameTime gameTime)
         {
-            if (state != GuildDialogState.Initial && _state == state)
-                return;
+            if (_guildDescription != _guildSessionProvider.GuildDescription && _guildSessionProvider.GuildDescription != "")
+            {
+                _guildDescription = _guildSessionProvider.GuildDescription;
+                SetState(GuildDialogState.Modify);
```
ethanmoffat, line 86: `SetState` as part of the update doesn't really make sense to me. IMO it should be the result of an action the user takes. Is guild dialog supposed to enter the `Modify` state when the description is changed?

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -229,5 +287,20 @@ private void SetStateIfGuildMember(GuildDialogState state)
 
             SetState(state);
         }
+
+        private void ShowChangeDescriptionMessageBox()
+        {
+            // TODO: Move max description length to config?
```
ethanmoffat, line 293: If this is a hard-coded constant in the vanilla client, I'd say make it a hard-coded constant in EndlessClient as well. No need for config on this.

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -151,8 +174,16 @@ private void SetState(GuildDialogState state)
                             SubText = _localizedStringFinder.GetString(EOResourceID.GUILD_CHANGE_YOUR_GUILD_DETAILS),
                             OffsetY = 45,
                         };
-                        modifyGuildItem.LeftClick += (_, _) => SetStateIfGuildMember(GuildDialogState.Modify);
-                        modifyGuildItem.RightClick += (_, _) => SetStateIfGuildMember(GuildDialogState.Modify);
+                        modifyGuildItem.LeftClick += (_, _) =>
+                        {
+                            _guildActions.GetGuildDescription(_characterProvider.MainCharacter.GuildTag);
```
ethanmoffat, line 179: If the character isn't in a guild this should probably not be sent to the server. Maybe call `SetStateIfGuildMember(GuildDialogState.Modify)`, and if the guild dialog's state was changed then send the packet?

EndlessClient/Dialogs/GuildDialog.cs
```diff
@@ -213,9 +244,36 @@ private void SetState(GuildDialogState state)
                         AddItemToList(disbandItem, sortList: false);
                     }
                     break;
+                case GuildDialogState.Modify:
+                    {
+                        ListItemType = ListDialogItem.ListItemStyle.Large;
```
ethanmoffat, line 249: Should this be changing the list item style to `Small`? All the list dialog items below are `ListItemStyle.Small`.

Also consider that changing the list item style within the dialog necessitates changing it *back*. @AstralSkies used a stack of dialog states in his PR, I think maybe it'd be worth doing the following:
1. Track a stack of DialogState/DialogStyle
2. Each call to `SetState` pushes the current state/style onto the stack
3. Each call to `BackAction` pops the previous state/style off the stack and sets to that state.

This way we don't have to maintain this complex web of which state goes to which style etc. etc. We just use what's on the stack.

EOLib/PacketHandlers/Guild/GuildTakeHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        public GuildTakeHandler(IPlayerInfoProvider playerInfoProvider,
+                                 IGuildSessionRepository guildSessionRepository)
+            : base(playerInfoProvider)
+        {
+            _guildSessionRepository = guildSessionRepository;
+        }
+
+        public override bool HandlePacket(GuildTakeServerPacket packet)
+        {
+            _guildSessionRepository.GuildDescription = packet.Description;
```
ethanmoffat, line 33: It's unclear to me if this description is the player's current guild description or if it's the description of a guild requested via the "guild lookup" function.

If this is a search result, it shouldn't be in the guild session repository state - guild session repository should only be the current player's guild info. Otherwise this is fine as-is, but it'd be nice to clarify the member name slightly so it's easier to tell whose guild this description belongs to.

ethanmoffat (changes requested):

EndlessClient/Dialogs/GuildDialog.cs
AstralSkies: https://github.com/AstralSkies/EndlessClient/blob/guild_testing/EndlessClient/Dialogs/GuildDialog.cs

Yeah my branch uses a stack to manage dialog states. I've linked it

EndlessClient/Dialogs/GuildDialog.cs
sorokya: It's how I got the description to re-render. Is there a better way?

EndlessClient/Dialogs/GuildDialog.cs
sorokya: This was more of a refresh the get the description to properly show

EndlessClient/Dialogs/GuildDialog.cs
ethanmoffat: What I'd probably do is track a reference to the guild description list item (in this PR it's line 257-260) as a member of this dialog and update the text of that list item directly when the guild description changes, rather than having a separate local string in this dialog and then needing to call additional code to actually refresh the string that's really displayed. Because this dialog has multiple states, you'll have to be smart about cleaning up the reference to that list item when the state changes and it's no longer valid; `Optional<T>` is your friend here.

I'll see if I can figure out how to code up a suggestion for the PR since I think that's something that's built in to github's PR flow

EOLib/PacketHandlers/Guild/GuildTakeHandler.cs
sorokya: This is the players guild description not a lookup

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Now tracking state/buttons/listitemtype per state and using stack

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Fixed

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Removed comment

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Now tracking the list item in a private optional field as suggested

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Field removed

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Now tracking state/buttons/listitemtype per state and using stack


Now tracking the list item in a private optional field as suggested

EndlessClient/Dialogs/GuildDialog.cs
sorokya: Changed style to `Small`

ethanmoffat (approved):

PR #367, opened 2024-07-28
Implement Player Freeze (Walk Open, Walk Close)

### LINKED ISSUE #350: Implement "freeze" walk packets
`WALK_OPEN` and `WALK_CLOSED` packets for freeze/unfreeze a character need to be handled.

miou-gh:
WalkValidationActions and AttackValidationActions currently exists. FaceValidationActions, EmoteValidationActions, and SitValidationActions could exist, but wasn't sure if it'd be appropriate to create for a single feature like this.

Closes #350


EOLib/Domain/Login/PlayerInfoRepository.cs
```diff
@@ -53,6 +57,8 @@ public sealed class PlayerInfoRepository : IPlayerInfoRepository, IPlayerInfoPro
 
         public bool PlayerHasAdminCharacter { get; set; }
 
+        public bool IsPlayerFrozen { get; set; }
```
ethanmoffat, line 60: `PlayerInfoRepository` and interfaces are largely intended to be a pre-game menu thing. Once the player is in game, `Character` should be used instead.

`Frozen` state should be moved to the `Character` class in EOLib. I think this would also eliminate a lot of churn in the diffs due to introducing the `IPlayerInfoProvider` dependency where there's already a `CharacterProvider`.

EOLib/PacketHandlers/Walk/WalkPlayerHandler.cs
```diff
@@ -11,10 +11,10 @@
```
ethanmoffat, line 11: Looks like nothing actually changed here, please reset to the content of the `master` branch

EndlessClient/Controllers/ArrowKeyController.cs
```diff
@@ -100,8 +104,11 @@ private bool CurrentDirectionIs(EODirection direction)
         }
 
         private void FaceOrAttemptWalk(EODirection direction)
-        {
-            if (!CurrentDirectionIs(direction) && _characterProvider.MainCharacter.RenderProperties.IsActing(CharacterActionState.Standing))
+		{
+			if (_playerInfoProvider.IsPlayerFrozen)
```
ethanmoffat, line 108: If `IsPlayerFrozen` is checked here, it'll never get to the point of validating the action in `AttemptToStartWalking`. I don't see any action being taken on `WalkValidationResult.Frozen` - is there supposed to be a status label message?

EOLib/Domain/Character/CharacterActions.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using AutomaticTypeMapper;
```
ethanmoffat, line 1: Looks like nothing actually changed here, please reset to the content of the master branch

EndlessClient/Controllers/ControlKeyController.cs
```diff
@@ -4,6 +4,7 @@
 using EndlessClient.Rendering.Character;
```
ethanmoffat, line 4: Looks like nothing actually changed here, please reset to the content of the master branch

EndlessClient/Controllers/FunctionKeyController.cs
```diff
@@ -11,6 +11,7 @@
 using EndlessClient.Rendering.Map;
```
ethanmoffat, line 11: Looks like nothing actually changed here, please reset to the content of the master branch

EndlessClient/Input/FunctionKeyHandler.cs
```diff
@@ -27,12 +31,17 @@ protected override Option<Keys> HandleInput()
                 if (IsKeyHeld(key))
                 {
                     var isShiftHeld = IsKeyHeld(Keys.LeftShift) || IsKeyHeld(Keys.RightShift);
-                    if (_functionKeyController.SelectSpell(key - Keys.F1, isShiftHeld))
+                    var isFrozen = _playerInfoProvider.IsPlayerFrozen;
+
+                    if (isFrozen)
+                        return Option.None<Keys>();
```
ethanmoffat, line 37: You need to decide where the frozen logic is going to be handled. There are a couple different layers to this:

1. Key "handlers", which detect monogame input events and call the appropriate controller methods. This essentially binds a key input to some meaningful game action.
2. Key "controllers", which orchestrate the logic associated with a meaningful game action.
3. Input "validators", which determine whether it's ok for the meaningful game action to take place. This is essentially a factoring out of logic that would normally reside in a controller.

I don't see this logic needing to be repeated in all three places. IMO the way to approach this would be to put the logic in the validators, and handle the validation result appropriately (status message or no-op depending on what the vanilla client does). I don't think it makes sense in the handlers, because those are concerned only with detecting monogame events, and I don't think it makes sense in the controllers because other validations are already encapsulated by the validators.

EndlessClient/Input/InputHandlerBase.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using EndlessClient.GameExecution;
```
ethanmoffat, line 1: Looks like nothing actually changed here, please reset to the content of the master branch

ethanmoffat (changes requested):

ethanmoffat (approved):

PR #368, opened 2024-07-31
Dotnet format

MrDanOak:
* add .editoconfig with sensible defaults
* run dotnet format on the solution
image
image
image


.editorconfig
```diff
@@ -0,0 +1,270 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
```
MrDanOak, line 1: as mentioned this is auto-generated and can be edited in visual studio GUI

.gitattributes
```diff
@@ -1,8 +1,8 @@
 # Auto detect text files and perform LF normalization
-* text=auto
+* text=auto eol=crlf
```
MrDanOak, line 2: this is required to get Linux to honour the source code's crlf encoding regardless of what Linux would like to do

.editorconfig
```diff
@@ -0,0 +1,270 @@
...
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
+indent_size = 4
+indent_style = space
+tab_width = 4
+
+# New line preferences
+end_of_line = crlf
+insert_final_newline = false
```
ethanmoffat, line 16: I think I actually want this `true` - I've been trying to add trailing newlines throughout the solution as I see them missing

.editorconfig
```diff
@@ -0,0 +1,270 @@
...
+
+#### .NET Coding Conventions ####
+
+# Organize usings
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = false
```
ethanmoffat, line 22: Seems like setting this to true will reduce a lot of noise in the PR?

ethanmoffat (commented):
Starting with these two suggestions - I think that'll reduce most of the code churn in the PR and aligns with my preferred style. I'd like to make introduction of a formatter as unintrusive as possible. It'll make it easier to see the impact of the other editorconfig settings if there isn't as much `using` reorder noise or newlines at the ends of files being deleted.

.editorconfig
MrDanOak: Yeah this has been done :) still a few where system using weren't first but way less of them now

.editorconfig
MrDanOak: done :)

ethanmoffat (approved):
Thanks so much for this, this will make my life significantly easier when reviewing PRs. I was hoping that the usings/file EOL were more consistent than they were - too bad this has to touch so many files, but seems like it's unavoidable. Guess we're ripping off the band-aid!

PR #369, opened 2024-07-31
Add Multi-Line Text Support for Rank Descriptions

sjbmcg:
- Modified the input handling logic.

### DIFF
--- a/EndlessClient/Dialogs/TextMultiInputDialog.cs
+++ b/EndlessClient/Dialogs/TextMultiInputDialog.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using EndlessClient.Content;
 using EndlessClient.Dialogs.Services;
 using EndlessClient.HUD.Chat;
+using EndlessClient.UIControls;
 using EOLib;
 using EOLib.Graphics;
 using Microsoft.Xna.Framework;
@@ -21,8 +22,8 @@ namespace EndlessClient.Dialogs
             Four,
             Five,
             FiveWithScroll,
+            NineWithScroll
         }
-
         public struct InputInfo
         {
             public string Label;
@@ -35,9 +36,15 @@ namespace EndlessClient.Dialogs
 
         private readonly Vector2 _bottomOverlayDrawPosition;
         private readonly Rectangle _bottomOverlaySource;
+        private readonly Vector2 _rightOverlayDrawPosition;
+        private readonly Rectangle _rightOverlaySource;
+        private readonly ScrollBar _scrollBar;
+        private readonly XNATextBox[] _inputBoxes;
+        private XNALabel[] _inputLabels;
+        private readonly DialogSize _dialogSize;
+        private readonly INativeGraphicsManager _nativeGraphicsManager;
 
-        private readonly IXNATextBox[] _inputBoxes;
-
+        private int _previousScrollOffset = -1;
         public IReadOnlyList<string> Responses => _inputBoxes.Select(x => x.Text).ToList();
 
         public TextMultiInputDialog(INativeGraphicsManager nativeGraphicsManager,
@@ -51,6 +58,10 @@ namespace EndlessClient.Dialogs
             : base(nativeGraphicsManager, isInGame: true)
         {
             _background = GraphicsManager.TextureFromResource(GFXTypes.PostLoginUI, 57);
+            _nativeGraphicsManager = nativeGraphicsManager;
+            _dialogSize = size;
+            Vector2 okButtonPosition;
+            Vector2 cancelButtonPosition;
 
             switch (size)
             {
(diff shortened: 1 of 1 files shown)


EndlessClient/Dialogs/TextMultiInputDialog.cs
```diff
@@ -154,9 +201,68 @@ protected override void OnDrawControl(GameTime gameTime)
...
+        protected override void OnUpdateControl(GameTime gameTime)
+        {
+            if (_dialogSize != DialogSize.NineWithScroll)
+            {
+                base.OnUpdateControl(gameTime);
+                return;
+            }
+            _scrollBar.UpdateDimensions(9);
+
+            if (_inputBoxes.Length > _scrollBar.LinesToRender)
```
ethanmoffat, line 224: I think this condition will always be true if we're already checking that `_dialogSize` is equal to `NineWithScroll`.

Also - `LinesToRender` is set to `5` in the constructor and never changed, so even considering the earlier `if` I'm pretty sure this will always be true.

EndlessClient/Dialogs/TextMultiInputDialog.cs
```diff
@@ -154,9 +201,68 @@ protected override void OnDrawControl(GameTime gameTime)
             _spriteBatch.Begin();
             _spriteBatch.Draw(_background, DrawAreaWithParentOffset, _backgroundSourceRectangle, Color.White);
             _spriteBatch.Draw(_background, DrawPositionWithParentOffset + _bottomOverlayDrawPosition, _bottomOverlaySource, Color.White);
+            _spriteBatch.Draw(_background, DrawPositionWithParentOffset + _rightOverlayDrawPosition, _rightOverlaySource, Color.White);
+
+            _spriteBatch.Draw(_background, DrawPositionWithParentOffset + _rightOverlayDrawPosition, _rightOverlaySource, Color.White);
```
ethanmoffat, line 206: Looks like you added this line a second time unintentionally

EndlessClient/Dialogs/TextMultiInputDialog.cs
```diff
@@ -154,9 +201,68 @@ protected override void OnDrawControl(GameTime gameTime)
...
+        // OnUpdateControl needs to solely run if Dialog.Nine is in use
+
+        protected override void OnUpdateControl(GameTime gameTime)
+        {
+            if (_dialogSize != DialogSize.NineWithScroll)
+            {
+                base.OnUpdateControl(gameTime);
+                return;
+            }
+            _scrollBar.UpdateDimensions(9);
```
ethanmoffat, line 222: This should be part of the initialization logic for the scrollbar in the constructor, rather than on every update. The number of items being scrolled doesn't change so `UpdateDimensions` doesn't need to be called more than once

EndlessClient/Dialogs/TextMultiInputDialog.cs
```diff
@@ -154,9 +201,68 @@ protected override void OnDrawControl(GameTime gameTime)
...
+                        currLabel.DrawPosition = new Vector2(currLabel.DrawPosition.X, 69 + relativeIndex * 24);
+                    }
+                    else
+                    {
+                        currBox.Visible = false;
+                        currLabel.Visible = false;
+                    }
+                }
+            }
+            else if (_inputBoxes.Any(_item => _item.Visible) || _inputLabels.Any(_item => !_item.Visible))
```
ethanmoffat, line 253: Since the above condition is always true, this `else if` branch can be removed entirely since it is never executed

EndlessClient/Dialogs/TextMultiInputDialog.cs
```diff
@@ -154,9 +201,68 @@ protected override void OnDrawControl(GameTime gameTime)
...
+            if (_dialogSize != DialogSize.NineWithScroll)
+            {
+                base.OnUpdateControl(gameTime);
+                return;
+            }
+            _scrollBar.UpdateDimensions(9);
+
+            if (_inputBoxes.Length > _scrollBar.LinesToRender)
+            {
+                for (int i = 0; i < _inputBoxes.Length; i++)
```
ethanmoffat, line 226: This approach looks correct but I don't like that this `OnUpdateControl` is executed unconditionally in its entirety every time (at least for `NineWithScroll` size).

What I'd do instead is track the previous scroll offset, and only execute this code if the previous scroll offset has changed. You'll see this pattern in just about every spot that uses a scroll bar in the source code - it reduces the amount of code executed from *all the code* to just the single condition.

ethanmoffat (changes requested):
This looks good, I appreciate you unwinding my attempt at a scroll bar API and using it successfully :) overall approach looks good, just a few notes in the update.

If you haven't verified it yet, please double-check that the existing functionality for the `Two` dialog size still works as expected. IIRC that's used for boards.

AstralSkies:
This should resolve the conflicts now.

ethanmoffat (approved):

PR #384, opened 2024-10-23
Implement guild bank deposit logic

ethanmoffat:


sorokya (approved):
Code looks good

EndlessClient/Subscribers/GuildEventSubscriber.cs
```diff
@@ -138,5 +138,13 @@ public void NotifyConfirmCreateGuild()
                 dlg.ShowDialog();
             });
         }
+
+        public void NotifyNewGuildBankBalance(int balance)
+        {
+            _sfxPlayer.PlaySfx(SoundEffectID.BuySell);
+
+            var dlg = _messageBoxFactory.CreateMessageBox(DialogResourceID.GUILD_DEPOSIT_NEW_BALANCE, $" {balance}");
```
sorokya, line 146: This should also have `goldName` after the value

sorokya (changes requested):
Need to add gold name after value in deposit dialog

PR #387, opened 2024-10-27
Implement 'Set Member Rank' and 'Update Guild Ranks' functions in GuildDialog

ethanmoffat:


ethanmoffat:
PR is borked, github screwed something up. Opening with a new PR reference.

PR #428, opened 2025-12-19
Camera system

do4k:


EndlessClient/HUD/Panels/InventoryPanel.cs
ethanmoffat, line 1: This change should be in a separate PR; it has nothing to do with the camera system. I'd also like some perf metrics that prove it's worth changing to this. IMO the existing way is much easier to understand

EndlessClient/Rendering/Effects/EffectRenderer.cs
ethanmoffat, line 1: ditto here. This looks like it addresses perf issues *and* a bug in how effects are drawn w.r.t. `OnTopOfCharacter`, so I'd like to take it, just not as part of a "camera system" PR.

EndlessClient/Rendering/Map/DynamicMapObjectUpdater.cs
```diff
@@ -31,6 +31,9 @@ public class DynamicMapObjectUpdater : IDynamicMapObjectUpdater
         private IMapFile _cachedMap;
         private List<MapCoordinate> _ambientSounds;
 
+        private readonly Dictionary<MapCoordinate, List<ICharacterRenderer>> _charactersByCoord = new();
+        private readonly List<ICharacterRenderer> _tempCharacterList = new();
```
ethanmoffat, line 35: Unused variable (surprised this isn't generating a warning)

EndlessClient/Rendering/MapEntityRenderers/MapItemLayerRenderer.cs
```diff
@@ -37,7 +37,7 @@ public override void RenderElementAt(SpriteBatch spriteBatch, int row, int col,
         {
             var items = _currentMapStateProvider.MapItems[new MapCoordinate(col, row)];
 
-            foreach (var item in items.OrderBy(item => item.UniqueID))
```
ethanmoffat, line 40: I think this is required to provide consistency to the order in which the map items are drawn. If it is removed, the click order and render order will be different. I apparently didn't capture the details in the commit that added it but I vaguely remember picking up items in a stack and it kinda just randomly grabbed one instead of taking the one on top.

EndlessClient/Rendering/NPC/NPCAnimator.cs
```diff
@@ -17,6 +17,9 @@ public class NPCAnimator : GameComponent, INPCAnimator
         private readonly Dictionary<int, (MapCoordinate Coord, EODirection Direction)> _queuedWalk = [];
         private readonly Dictionary<int, EODirection> _queuedAttack = [];
 
+        private readonly List<int> _npcsDoneWalking = [];
```
ethanmoffat, line 20: What does making these into fields accomplish? Less GC passes?

EndlessClient/HUD/Panels/InventoryPanel.cs
ethanmoffat: Ok I see this applies to a number of the files in this PR - if you want to do a perf fixes PR separately that's fine, please scope this one just to what is needed for the camera system.

ethanmoffat:
I'd love to see some data about before/after performance for the camera.

EndlessClient/Rendering/GridDrawCoordinateCalculator.cs
```diff
@@ -160,6 +165,22 @@ private Vector2 GetMainCharacterWalkAdjustOffsets()
...
+            var currentHeight = _clientWindowSizeProvider.Height;
+            var isResizable = _clientWindowSizeProvider.Resizable;
+
+            if (!_cacheValid || _cachedViewportWidthFactor != currentWidth / 2)
+            {
+                _cachedViewportWidthFactor = currentWidth / 2;
+                _cachedViewportHeightFactor = isResizable
+                    ? currentHeight / 2
+                    : currentHeight * 3 / 10 - 2;
+                _cacheValid = true;
```
ethanmoffat, line 181: When is `_cacheValid` reset to false? Does the window resize operation create a new `GridDrawCoordinateCalculator`?

EndlessClient/Rendering/GridDrawCoordinateCalculator.cs
ethanmoffat, line 1: I've had the idea of a camera system in the back of my mind for years. One of the goals was to delete this class completely and rely on the camera system to know where everything should be rendered.

Would it be possible to completely replace `GridDrawCoordinateCalculator` and `RenderOffsetCalculator`?

EndlessClient/Rendering/Camera2D.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        /// <summary>
+        /// Forces a camera update on the next frame.
+        /// </summary>
+        void ForceUpdate();
+
+        /// <summary>
+        /// Converts world coordinates to screen coordinates.
+        /// </summary>
+        Vector2 WorldToScreen(Vector2 worldPosition);
```
ethanmoffat, line 177: Which of these are actively used? IMO the interface shouldn't be exposing stuff that isn't needed (although I can see how this stuff would come in handy if the `ICamera2D` becomes the source of truth for rendering coordinates rather than having `GridDrawCoordinateCalculator`)

EndlessClient/Rendering/Camera2D.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        /// <summary>
+        /// Gets the camera offset for the main character (cached).
+        /// This replaces the need to constantly recalculate GetMainCharacterOffsets().
+        /// </summary>
+        public Vector2 GetMainCharacterOffset()
+        {
+            var props = _characterProvider.MainCharacter.RenderProperties;
+            return new Vector2(
+                _renderOffsetCalculator.CalculateOffsetX(props),
```
ethanmoffat, line 121: This appears to be less cached than the doc comment would lead me to believe

EndlessClient/Rendering/Camera2D.cs
```diff
@@ -0,0 +1,194 @@
...
+        private Matrix _transformMatrix;
+        private bool _isDirty = true;
+        private int _lastCharacterX = -1;
+        private int _lastCharacterY = -1;
+        private int _lastCharacterOffsetX = 0;
+        private int _lastCharacterOffsetY = 0;
+        private int _lastViewportWidth = -1;
+        private int _lastViewportHeight = -1;
+
+        public Vector2 Position => _position;
```
ethanmoffat, line 29: I'm missing where the Position/TransformMatrix are actually used by the code - seems like these should be propagated all over the place (i.e. SpriteBatch.Draw calls?)

EndlessClient/Rendering/Camera2D.cs
```diff
@@ -0,0 +1,194 @@
...
+                _lastCharacterOffsetX = currentOffsetX;
+                _lastCharacterOffsetY = currentOffsetY;
+                _lastViewportWidth = currentWidth;
+                _lastViewportHeight = currentHeight;
+
+                UpdatePosition();
+                _isDirty = true;
+            }
+
+            if (_isDirty)
```
ethanmoffat, line 73: I would have expected `_isDirty` to guard more of the calculations that are done. Is computing a translation matrix that expensive?

EndlessClient/Rendering/Camera2D.cs
```diff
@@ -0,0 +1,194 @@
...
+        public Vector2 ScreenToWorld(Vector2 screenPosition)
+        {
+            return screenPosition + _position;
+        }
+
+        /// <summary>
+        /// Gets the camera offset for the main character (cached).
+        /// This replaces the need to constantly recalculate GetMainCharacterOffsets().
+        /// </summary>
+        public Vector2 GetMainCharacterOffset()
```
ethanmoffat, line 117: This method doesn't feel like it belongs here, especially since it just forwards to `RenderOffsetCalculator`

ethanmoffat (changes requested):

